PR #82, opened 2014-06-25
Fixing an url initialization issue.

ericbulloch:
Using reverse instead of reverse_lazy was invoking the django urlresolvers
too early. Other apps didn't finish initializing and so apps started to
brake. reverse_lazy fixed the issue.


camilonova:
@ericbulloch want to take a look into it?

PR #108, opened 2014-11-10
Adding support for AXES_ONLY_USER_FAILURES checking to ignore IP address...

lip77us:
I needed to be able to lock out users independent of their IP address and their user_agent.  This implementation adds a setting AXES_ONLY_USER_FAILURES which when true will lock the account no matter which user_agent or which IP the request come from.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -434,7 +434,7 @@ def check_request(request, login_unsuccessful):
     user_lockable = is_user_lockable(request)
     # no matter what, we want to lock them out if they're past the number of
     # attempts allowed, unless the user is set to notlockable
-    if failures > FAILURE_LIMIT and LOCK_OUT_AT_FAILURE and user_lockable:
+    if failures >= FAILURE_LIMIT and LOCK_OUT_AT_FAILURE and user_lockable:
         # We log them out in case they actually managed to enter the correct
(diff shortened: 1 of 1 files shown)


camilonova:
@lip77us can you tell why are you doing f63dee6 ?

lip77us:
Sure.  If I have set the limit to lockout after 3 failures, The screen 
was not jumping to the "your are locked out" screen until the 4th 
attempt.  I wanted to be able to tell users they were locked out as soon 
as they failed their 3rd attempt.

camilonova:
We'll the purpose of the lockout message is when you exceed the allowed number you'll see it. And we are not encouraging the users to see it, it's only as a last resort self defense mechanism.

I guess you could set the limit to 2 failures, so they will see the screen in their 3rd attempt.

lip77us:
Thats fine.  I will probably still keep the feature on my branch.  My 
issue was that on the 4th attempt, they may have actually gotten the 
password correct, but we tell them they are now locked out.  I preferred 
to tell them they are locked out when they are locked out so that they 
can contact an administrator to get unlocked.  My client test team was 
having a problem with the way it was working.

camilonova:
Thank you, I'm closing it for now.

PR #118, opened 2015-03-20
Fixed create_new_failure_records(), added feature to lock out by USER+IP

Oik741:
Removed unnecessary separating of one failure attempt on two AccessAttempt objects(according to current code it's not required).
Added optional feature to lock out account login from particular IP.


camilonova:
@Oik741 Thank you sir.

PR #119, opened 2015-04-29
Fixes sending the user_locked_out signal

### LINKED ISSUE #94: Not sending lockout signal
The library doesn't send the lockout signal. Tried both with decorator and middleware.
Debugging check_request, it does not pass through it when locking out the user, maybe the control
```python
if is_already_locked(request):
```
in watch_login prevents the signal sending by returning directly?
Or am I missing something else?

toabi:
It also adds a test for that. Fixes #94.


aclark4life:
w00t, thanks

camilonova:
Thanks

PR #120, opened 2015-06-23
properly use username and IP when pulling attempts. add test and easier testing command.

SteveByerly:
The latest setting didn't work the way I expected. It seems like it was supposed to allow one username to try to log in even if another user from the same IP was already locked. In practice, it kept making new AccessAttempt records instead of upping the failures_since_start field of the existing one.

I added a test to make sure switching the setting from False to True still works. I also couldn't find an easy way to test, so I added the manage.py file to the project root.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -238,10 +238,12 @@ def _get_user_attempts(request):
             ip_address=ip, username=username, trusted=True
         )
 
-    if not attempts and not LOCK_OUT_BY_COMBINATION_USER_AND_IP:
+    if not attempts:
         params = {'ip_address': ip, 'trusted': False}
         if USE_USER_AGENT:
             params['user_agent'] = ua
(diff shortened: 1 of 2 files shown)


camilonova:
I don't think adding a ```manage.py``` file is a good idea. Can you only leave the test?

SteveByerly:
Why is not adding the manage.py a good idea? I'm not super experienced with testing Django, but how else do you run the tests for the app? (serious questions)

I have an update to my test that will ensure that users with a different username are still able to log in, so I'd like to send another PR anyway - but I wanted to get some info before removing the manage.py.

Thanks!

camilonova:
@SteveByerly you just break the tests, can you take a look? https://travis-ci.org/django-pci/django-axes/builds/68053809

camilonova:
@SteveByerly I just reverted the changes, can you please take a look at the tests and send it again?

PR #123, opened 2015-07-06
Add username to LOCKOUT_TEMPLATE template context

daleobrien:
So I can use their username in the lockout template, e.g. something like this with the template,

 <p>There have been too many failed login attempts for {{username}}. Please email <a href='mailto:unlockme@example.com?subject=Account reset for "{{usernamel}}"&body=Hi, Please unlock my account. Thanks'>unlockme@example.com</a> to unlock the account.</p>


camilonova:
Thank you.

PR #125, opened 2015-08-31
Fix #_get_user_attempts to include username when filtering AccessAtte…

AFioca:
…mpts if AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP is True


camilonova:
Thanks

Issue #73, opened 2014-04-25
Possible Denial of Service with POST request data


peterkuma:
django-axes logs POST data of login requests in full. This allows a potentialy large amount of data to be logged in the database.

How to reproduce:

    dd if=/dev/urandom of=bigfile bs=1M count=1
    curl -F 'file=<bigfile' http://localhost:8000/admin/ # See note.

The model instance in AccessAttempts now contains the full content of the file. This can be misused to fill the disk space, or slow down the browser when viewing the model instance page (esp. Firefox does not handle it very well).

django-axes should only log a limited amount of POST data, e.g. 1 kB.

**Note:** For this to work, CSRF needs to be disabled, e.g. with:

example/settings.py:

    MIDDLEWARE_CLASSES = (
        ...
        'middleware.DisableCSRF',
    )

example/middleware.py:

    class DisableCSRF(object):
        def process_request(self, request):
            setattr(request, '_dont_enforce_csrf_checks', True)

camilonova:
Closing this in favor of #127

PR #129, opened 2015-10-09
Fixes #127 & #73 - Improved the implementation of query2str

### LINKED ISSUE #127: Improvements to axes.decorators.query2str
Scanning through the code, I noticed that this method had the following note: `The length of the output is limited to max_length to avoid a DoS attack.`

If performance is a legitimate concern in this method due to potential DoS attack, I noticed a couple of issues with the implementation.

I ran a few tests using the Python `timeit` module. Using a dictionary containing 10,000 items and iterating over the call to query2str() 1,000 times I got an average over 3 runs of 5.70709357833645 seconds. By removing the `[:max_length]` limiter, I got an average over 3 runs of 5.64547695267053 seconds. This difference is somewhat negligible, and could be written off to me running this on a busy machine.

Now, for one first-pass optimization, I moved the `[:max_length]` limiter, changing the line from `return '\n'.join(kvs)[:max_length]` to `return '\n'.join(kvs[:max_length])`; limiting the number of items before the join. With this change, I got an average of 5.573779275335255 seconds over 3 runs. But, there is one more big optimization to be made here: list comprehensions.

Switching to a list comprehension, I was able to get an average of 5.097600340998421 seconds over 3 runs. Now, while 0.6 seconds doesn't seem like a whole lot, keep in mind that this is one thousand iterations with dictionaries containing ten thousand keys. Increasing the number of iterations or size of the dictionary, this gap would widen significantly. If this method is a true concern, I think this would be a worthwhile change.

The test code follows:

```python
import timeit

setup = """
import six

PASSWORD_FORM_FIELD = 'password'

def query2str(items, max_length=1024):
    kvs = []
    for k, v in items:
        if k != PASSWORD_FORM_FIELD:
            kvs.append(six.u('%s=%s') % (k, v))

    return '\\n'.join(kvs)[:max_length]

dict = {x: x for x in range(0, 10000)}
"""

timeit.timeit('query2str(dict.items())', setup, number=1000)


setup = """
import six

PASSWORD_FORM_FIELD = 'password'

def query2str(items, max_length=1024):
    return '\\n'.join([six.u('%s=%s' % (k, v)) for k, v in six.iteritems(items) if k != PASSWORD_FORM_FIELD][:max_length])
dict = {x: x for x in range(0, 10000)}
"""
timeit.timeit('query2str(dict)', setup, number=1000)
```

Note that this also moves the iteration inside `query2str()` method and uses the Python 2/3 compatible `six.iteritems()` instead of just calling `request.POST.items()` and `request.GET.items()` outside.

tarkatronic:


tarkatronic:
Most of the failures in the CI build were caused by the call to `six.u()`, which I brought over from the previous implementation. If I simply remove that, the test suite passes on my machine in both Python 3.4.3 and 2.7.9. Django 1.4.15 w/Python 3.4, I just forgot to exclude.

Is there a specific reason that call was needed? I'm not entirely understanding it's purpose from the six docs. I see that call was added here: https://github.com/django-pci/django-axes/commit/c90ee0562afd4598ca7e20462bdeeca0dc02e386#diff-98c00a3d39f41b919aeb159f2b5dd247 but I don't really see how it is necessary.

tarkatronic:
It looks like the introduction of using u'%s=%s' was here: https://github.com/django-pci/django-axes/commit/6cc216c14a8c60ed855eb733e9d87331493b568f#diff-98c00a3d39f41b919aeb159f2b5dd247

I'm not entirely convinced it's necessary.

axes/decorators.py
```diff
@@ -160,15 +160,11 @@ def query2str(items, max_length=1024):
...
     """
 
-    kvs = []
-    for k, v in items:
-        if k != PASSWORD_FORM_FIELD:
-            kvs.append(six.u('%s=%s') % (k, v))
-
-    return '\n'.join(kvs)[:max_length]
+    return '\n'.join(['%s=%s' % (k, v) for k, v in six.iteritems(items)
+                      if k != PASSWORD_FORM_FIELD][:int(max_length/2)])[:max_length]
```
jdunck: What's up with this "int(max_length/2)" ?

tarkatronic: That's a minor optimization to limit the number of elements passed to the join(), in case of a GET or POST payload with an excessive number of elements. It's a little shortcut, since the end result is truncated to `max_length` anyways. Actually, this could probably even be `int(max_length/4)`. I'm having trouble properly articulating it right now (brain malfunction) but, given that the output is in the format `k=v\n`, the most elements which could fit into the eventual output is `max_length/4` (4 characters per element).

PR #138, opened 2015-12-17
Add missing db indexes for CommonAccess model

### LINKED ISSUE #137: DB indexes missing on AccessAttempt
The model `AccessAttempt` does not contain any DB indexes. If this table contains a lot of entries then the login performance is very slow. I saw it already happen with 50K+ entries.

The following fields should have an index to prevent performance issues:
- user_agent
- ip_address
- username
- trusted

See `decorators.py:_get_user_attempts`: https://github.com/django-pci/django-axes/blob/master/axes/decorators.py#L229-L235

Schweigi:
Fixes #137.


camilonova:
@Schweigi Thank you

PR #141, opened 2016-01-29
Reverse Proxy Whitelist fix

HageMaster3108:
The IP whitelist check in get_ip() asssumes that AXES_IP_WHITELIST is configured and dies if it is "None" (minor issue).

Fixed by using ip_in_whitelist(ip).


camilonova:
Thank you

PR #144, opened 2016-03-11
Made migrations py3 compatible

### LINKED ISSUE #142: Migrations in Python3
In the old migrations some fields were prefixed with b'
In a Python3 environment this resulted into:

    You have unapplied migrations; your app may not work properly until they are applied.
    Run 'python manage.py migrate' to apply them.

Because Python3 wants to remove those b' prefixes.
I changed it in my fork, however, I suspect that on a Python2.x environment you'll get the same error message if you would use my Py3 fork with a Django version lower than 1.7

https://code.djangoproject.com/ticket/22564

ideallical:
Fixes #142


camilonova:
Thank you

Issue #126, opened 2015-09-15
axes_reset command errors in django 1.8


Bartvds:
We're testing django-axes 1.5.0 from PyPi in a project upgraded to django 1.8.4, but when we try to reset an IP address in a management command we get this error:

````
$ ./manage.py axes_reset 123.123.123.123
usage: manage.py axes_reset [-h] [--version] [-v {0,1,2,3}]
                            [--settings SETTINGS] [--pythonpath PYTHONPATH]
                            [--traceback] [--no-color]
manage.py axes_reset: error: unrecognized arguments: 123.123.123.123
````

I fairly sure this is related to this: https://docs.djangoproject.com/en/1.8/releases/1.8/#management-commands-that-only-accept-positional-arguments

Maybe this is because the args command variable is an empty string? https://github.com/django-pci/django-axes/blob/master/axes/management/commands/axes_reset.py#L8

int-ua:
Confirming this for axes `1.6.0` with Django `1.8.11`

PR #145, opened 2016-03-11
Fixing #126, possibly breaking compatibility with Django<=1.7

int-ua:
It should be possible to implement compatibility but I'm not interested in this at all.


int-ua:
#126

camilonova:
@aclark4life seems we can't forget projects using django < 1.7 what do you think?

aclark4life:
I would say we *can* drop 1.7 now, see: https://www.djangoproject.com/download/#supported-versions

camilonova:
Thanks

Issue #107, opened 2014-11-03
middleware problem


codefisher:
The middleware will redecorate the `auth_views.login` function on every single request.  That is why you need the `if func.__name__ == 'decorated_login':` inside the watch_login.  You should rather do this check inside the middleware, or better, add the watch_login wrapper somewhere else.  I don't see a reason it needs to be in middleware at all?

teeberg:
We just ran into "maximum recursion depth" errors in testing for that reason. I've proposed a PR to fix this.

camilonova:
Thank you

PR #149, opened 2016-04-22
Decorate auth_views.login only once

teeberg:
During testing, the middleware is created (i.e. `__init__` is called) for every new request through the django test client. This PR adds a check, as suggested in #107, to not re-decorate `auth_views.login` over and over it if it's already been decorated.


camilonova:
Thank you

Issue #151, opened 2016-04-27
Some 192.-prefix addresses are non-private, but still get turned into 127.0.0.1


SilasX:
While using, we had a user from a non-private IP address with a 192 prefix (i.e. in 192.0.0.0/8).  They showed up in this module's admin logs as coming from 127.0.0.1 even though they came from an external, public IP.

However, per [the RFC](https://en.wikipedia.org/wiki/Private_network#Private_IPv4_address_spaces), it's only the 192.168 prefix (192.168.0.0/16) that is private.  [This line](https://github.com/django-pci/django-axes/blob/master/axes/decorators.py#L102) in the code is what is classifying the broader set that way, but the `'192.'` should probably be `'192.168.'`.

camilonova:
Should be fixed with #154

PR #154, opened 2016-04-27
Set IP public/private classifier to be compliant with RFC 1918.

SilasX:
Per #151. Also broke out the "private IP check" as its own function (refactoring related checks) with tests to ensure that not all 172- and 192- prefix IPs are classified as public.  Resubmit of previous PR #153 because a function was mis-documented.

Preserved prefix checking for determining private/public rather than [these approaches](http://stackoverflow.com/questions/691045/how-do-you-determine-if-an-ip-address-is-private-in-python) in order to preserve similarity to Axes module's approach.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -97,20 +97,32 @@ def is_valid_ip(ip_address):
     return valid
 
 
+def is_valid_public_ip(ip_address):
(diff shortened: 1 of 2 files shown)


camilonova:
Very nice, thank you.

PR #157, opened 2016-05-12
Missing migration

smeinel:
Hi, it seems that there's a migration missing from axes at the moment. Here's the migration I get after invoking `manage makemigrations axes`.


camilonova:
Thank you

PR #163, opened 2016-06-01
Fixes #162: UnicodeDecodeError on pip install

### LINKED ISSUE #162: UnicodeDecodeError
Hi,

I'm encountering an issue when trying to setup django-axes. When invoking command pip3 install django-axes, I'm having the following error:

> Downloading django-axes-1.6.1.tar.gz
>     Complete output from command python setup.py egg_info:
>     Traceback (most recent call last):
>       File "<string>", line 1, in <module>
>       File "/tmp/pip-build-eepf_eck/django-axes/setup.py", line 15, in <module>
>         open("CHANGES.txt").read()),
>       File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
>         return codecs.ascii_decode(input, self.errors)[0]
>     UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 545: ordinal not in range(128)

It seems there is a character in CHANGES.txt which does not belong to standard ascii table (probably russian characters). Could you please fix it by forcing utf-8 decoding?

joeribekker:


camilonova:
Thank you

PR #167, opened 2016-06-20
Fix ISO8601 time duration format.

mullakhmetov:
Missed 'T' before time part of iso8601 formatted time duration.


camilonova:
Thank you

PR #169, opened 2016-07-07
Add `default_app_config` so you can just use `axes` in `INSTALLED_APPS`

vdboor:
By adding `default_app_config` to `__init__.py`, there is no need to change the `INSTALLED_APPS` entry when upgrading to axes 2.0. I think this would make life much easier for developers with many projects. - upgrading axes would involve just removing the middleware class.

It also ensures that the login view is always patched.


camilonova:
Thank you

PR #182, opened 2016-08-10
Added possibility to disable access log

svenhertle:
The access log is not used for the blocking of IPs and users. I added a option to disable the access log since I do not want to have these data.

Due to a circular dependency when adding the new option DISABLE_ACCESS_LOG and including it in signals.py, I moved all settings to the file settings.py.


camilonova:
Thank you. Please add a test where you can see an unsuccessful attempt is logged if the setting is True

svenhertle:
This was indeed not the case until now. I fixed that, added the test and formulated the docs more clear.

camilonova:
Looks good to me. Thank you.

PR #184, opened 2016-08-17
Fix for IIS used as reverse proxy

Dmitri-Sintsov:
When IIS is used as reverse proxy, it adds port number to 'HTTP_X_FORWARDED_FOR' header causing 'axes_accesslog' table insertion error for field 'ip_address'.

By default the setting is turned off.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -33,6 +33,9 @@ def get_ip(request):
     if BEHIND_REVERSE_PROXY:
         ip = request.META.get(REVERSE_PROXY_HEADER, '').split(',', 1)[0]
         ip = ip.strip()
+        if REVERSE_PROXY_XFF_CLIENT_PORT:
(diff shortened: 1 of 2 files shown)


camilonova:
Thank you. I guess we can use a safe strategy here to remove the port if present. I'll make the change in another commit.

Dmitri-Sintsov:
I was worrying that IPv6 addresses could contain ':' characters while not being the port separator. But I do not know whether Postgres field of 'inet' type would accept both IPv6 and IPv4. Our networks are IPv4. Also it seems that IIS is non-standard to include client port in XFF header. I know that it is unusual to host Django projects in Windows / IIS but unfortunately my customer has Windows Server based infrastructure including backup solutions, they do not want to move to Linux.

camilonova:
Can you please check the IPv6 so we can think for a solution that covers it?

I think first we can check if is an IPv6 address, if not we can remove the ':'

PR #187, opened 2016-08-20
Use ipaddress to detect IPv6 addresses in decorators.get_ip()

Dmitri-Sintsov:
Fixes IIS supplying port number in X_FORWARDED_FOR reverse proxy header, which is incompatible to Postgres 'inet' field type.

Uses either 3.3+ built-in module or 2.x backport from requirements.txt.


Dmitri-Sintsov:
Sorry but my project uses Python 3.4 so I cannot check in 2.7 and travis does not use ipaddress compatibility module from requirements.txt.

camilonova:
We are not going to add an external dependency just for this.

I'm sure Django already provides something around it. Please check.

Also please consider we are not supporting python 3.4.

PR #188, opened 2016-08-22
Removed ipaddress module dependency.

Dmitri-Sintsov:
Uses socket.inet_pton() to detect IPv6 address to remove client port from IIS 'X_FORWARDED_FOR' HTTP header which causes invalid 'inet' type value error when inserting into Postgres database table.


camilonova:
Thank you

PR #190, opened 2016-09-06
Made the signal race condition safe.

Minkey27:
- Removed unnecessary import.
- Made the changing of access_logs in signals.py race condition safe.


camilonova:
Thanks

PR #191, opened 2016-09-22
Added AXES_ONLY_USER_FAILURES to support only looking at the user ID …

lip77us:
…and not the IP address.  I needed to add this for offices that use the same IP.  One user was locking the whole office out of my application.  Tests updated as well.

I just branch off of latest so this should be able to be merged.

All the tests including my new added tests ran without issue.


camilonova:
Thank you

PR #195, opened 2016-09-27
Fix bug when using the optional IP parameter

EvaMarques:
When the IP parameter is used the first element of kwargs needs to be skipped because its value is the string 'ip'.


camilonova:
Good catch. Thank you.

PR #196, opened 2016-10-24
Reset lockouts for user

vladimirnani:
Find this useful. Sometimes we need to reset lockout for specific user, maybe after phone call.


camilonova:
Thank you

PR #200, opened 2016-11-04
Added proper DISABLE_ACCESS_LOG. (With succeeding test suite)

Minkey27:
Hello,

I've added proper disabling of the accesslog. As far as I know this won't hurt the functionality of django-axes. (If I'm wrong, please correct me).

Also added DISABLE_SUCCESS_ACCESS_LOG which used to be DISABLE_ACCESS_LOG.
I have changed this because the setting name was not doing what you would expect it to do.

By changing the code this way, the "standard" behaviour is back to what it used to be.

(With succeeding test suite this time)


camilonova:
Thank you

PR #203, opened 2016-11-10
Update configuration.rst

benkonrath:
I found the language in the `AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP` option confusing so I've updated it. I took a bit of guess as to the exact meaning so please check if my interpretation is ok. Thanks.


camilonova:
Thank you

PR #207, opened 2016-11-17
Only look for lockable users on a POST.

### LINKED ISSUE #205: get() returned more than one User -- it returned 2! (Multiple None user names)
Attempting to use axes with combination user and ip has resulted in the entire site being unreachable, every single page simply showing the plain message: `get() returned more than one User -- it returned 2!`

This is before a user has actually logged into django at all. Our system allows for multiple `None` usernames as not every user actually needs to be able to log in, while it appears axes has (admittedly fairly) assumed that there will never be more than one user with a blank username.

I haven't dug any further into this, if there is some configuration that will allow axes to work with our system then that would be wonderful.

schinckel:
Resolves #205.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -94,6 +94,9 @@ def is_user_lockable(request):
     If so, then return the value to see if this user is special
     and doesn't get their account locked out
     """
+    if request.method != 'POST':
+        return False
+
     try:
         field = getattr(get_user_model(), 'USERNAME_FIELD', 'username')
         kwargs = {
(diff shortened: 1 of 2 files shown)


schinckel:
Not quite sure how to go about writing a test for #205, since it requires DB changes. However, I did test it against a database with NULLABLE usernames, and it allows doing a GET, and then correctly locks out a user when conditions are met.

schinckel:
I'm not actually sure this will wash, as I believe subsequent attempts to do a GET from a page that is already locked will be permitted: I suspect this is not the desired behaviour.

camilonova:
Thank you

Issue #206, opened 2016-11-15
Empty IP logged when using IPv4 behind reverse proxy


teeberg:
When running behind a reverse proxy, [this line](https://github.com/jazzband/django-axes/blob/ef3d527bee81e10455f73e57aaea13a222c9b792/axes/decorators.py#L43)
```python
ip = request.META.get(REVERSE_PROXY_HEADER, '').split(',', 1)[0]
```
correctly gets my external IP, e.g. `208.67.222.222`. Since it's not IPv6, [this code](https://github.com/jazzband/django-axes/blob/ef3d527bee81e10455f73e57aaea13a222c9b792/axes/decorators.py#L57) runs:
```python
        if not is_ipv6(ip):
            # Fix for IIS adding client port number to 'HTTP_X_FORWARDED_FOR' header (removes port number).
            ip = ''.join(ip.split(':')[:-1])
```
and sets `ip = ''`, which is then returned, resulting in this logged access attempt:<br>
![image](https://cloud.githubusercontent.com/assets/199071/20320696/a3114f8c-ab27-11e6-8e37-a0f09f267c42.png)

I'm not using IIS.

teeberg:
This was introduced in https://github.com/jazzband/django-axes/commit/50b9d97281b42f9508c69089e1f026477468eb9d, so it looks like it affects all versions since 2.2.1.

Issue #171, opened 2016-07-27
Make sure header parsing is correct


IvanAnishchuk:
Details in the inline comment https://github.com/jazzband/django-axes/commit/0239e173e00d7a1e7cc60d22d09808cd262f0f7d#commitcomment-18422665

camilonova:
Seems like a good approach, please make a PR and tests with your desired support.

Please consider that I should not be able to send the IP address in the headers before the proxy server. Otherwise I could create a script that changes the IP address every time I make a request.

IvanAnishchuk:
If you use nginx, you can rely on `realip` module, set a list of trusted proxies, and use `x-real-ip` header populated by `realip`.

Without a smart reverse proxy, that logic can be replicated by having a list of trusted proxies in settings, checking if remote_addr is one of them, and filtering `x-forwarded-for` and using the last entry left there instead of the first. Because the first can be supplied by a third party and hence easily spoofed.

To be clear, I'm not interested in doing anything here, axes is unfortunately unusable in my project, don't expect code from me. But its default behavior seems like too much of an error, making it susceptible to spoofing and all, so I thought I'd stop by and mention it.

camilonova:
Could you elaborate a test script to show the error? With it someone could fix it.

PR #209, opened 2016-11-21
Fix and add tests for IPv4 and IPv6 parsing from proxy X-Forwarded-For headers

### LINKED ISSUE #186: Add tests for BEHIND_REVERSE_PROXY
Add tests to check when BEHIND_REVERSE_PROXY is True

aleksihakli:
**Note:** This patch does not fix IPv6 parsing with ports

IP parsing is broken because of a broken past IIS patch for port resolving, and does not function correctly on e.g. AWS Elastic Beanstalk.

Standard formatted IPv4 addresses are reduced to empty string in [current version](https://github.com/jazzband/django-axes/blob/master/axes/decorators.py#L57):

    $ python
    Python 3.5.2 (v3.5.2:4def2a2901a5, Jun 25 2016, 22:18:55) [MSC v.1900 64 bit (AMD64)] on win32
    Type "help", "copyright", "credits" or "license" for more information.
    >>> ip = '192.168.1.1'
    >>> ip = ''.join(ip.split(':')[:-1])
    >>> ip
    ''

The correctly functioning port parsing is of format:

    >>> ip = '192.168.1.1'
    >>> ip = ''.join(ip.split(':', 1)[0])
    >>> ip
    '192.168.1.1'

This behaviour seems to be because of duplicate and conflicting patch entires for the `get_ip(request)` function.

This patch fixes IP parsing for normally formatted IPv4 addresses with ports, and IPv6 addresses without ports. (It does not mutilate IPv6 addresses, but keeps the current functionality). It cleans up the internal implementation of get_ip and adds a test set for the `get_ip(request)` function.

To accommodate the changes, test runner setup had to be refactored, and different settings added for axes setups running behind proxies. This only affects internal APIs and does not modify existing functionality.

This patchset should fix #186 and #206 and can be merged with #171.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ env:
 install:
 - pip install -q $DJANGO coveralls
 script:
-- coverage run --source=axes runtests.py
+- coverage run -a --source=axes runtests.py
+- coverage run -a --source=axes runtests_proxy.py
+- coverage run -a --source=axes runtests_proxy_custom_header.py
 - coverage report
 after_success:
 - coveralls
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -37,15 +37,18 @@ def is_ipv6(ip):
 
 
 def get_ip(request):
-    ip = request.META.get('REMOTE_ADDR', '')
+    """Parse IP address from REMOTE_ADDR or
+    AXES_REVERSE_PROXY_HEADER if AXES_BEHIND_REVERSE_PROXY is set."""
 
     if BEHIND_REVERSE_PROXY:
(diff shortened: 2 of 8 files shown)


camilonova:
Thank you

Dmitri-Sintsov:
Was it really necessary to remove fix for IPv6 parsing with ports? IIS still uses such address in the X-Forwarded-For header. Fixing one trouble while removing the useful functionality.

aleksihakli:
Well, not absolutely necessary, but it did fix a honest bug for more people than it broke things for. 

If you wish to have the IIS header parser in upstream and new releases, we can merge PRs and make new releases rather quickly.

Thank you for your understanding and appreciation for the voluntary maintenance and R&D that has gone into this project.

PR #216, opened 2017-01-12
Log failed access attempts regardless of settings

### LINKED ISSUE #212: `AXES_ONLY_USER_FAILURES=True` doesn't seem to cause lockout
Issue: When `AXES_ONLY_USER_FAILURES=True`, no `AccessAttempt` objects are created and therefore never actually locked out. See: https://github.com/jazzband/django-axes/blob/ca55a6ef16acc2f8f5f0501de4ef8f0d2ccd6129/axes/decorators.py#L431

Settings:
`AXES_ONLY_USER_FAILURES=True`

Reproducible steps:
1. Attempt to login with X username 3 times with a wrong password (default `AXES_LOGIN_FAILURE_LIMIT`)
2. Upon third attempt you see the lockout screen/message
3. Either refresh the page, or go back to admin URL
4. You can continue to try and login for an infinite(?) number of times for username X
5. You can then login with username X using the correct password

jimr:
Fixes #212

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -433,19 +433,20 @@ def create_new_failure_records(request, failures):
     ua = request.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
     username = request.POST.get(USERNAME_FORM_FIELD, None)
 
-    # record failed attempt from this IP if not AXES_ONLY_USER_FAILURES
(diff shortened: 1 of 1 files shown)


camilonova:
Thank you

jimr:
No problem, thanks for merging. Do you think you'll push out a bugfix release to PyPI any time soon?

PR #225, opened 2017-04-06
Fixed #224 -- Add AXES_NUM_PROXIES setting

### LINKED ISSUE #221: Always blocking the user until he solves the captcha
Currently, if a user tries to login for 3 times and fails to, he/she will be blocked. If he tries the fourth time with correct credentials, they will be successfully logged in.

I am trying to block a user even if he tries to login with the correct credentials and always redirect him/her to a captcha page. The `AccessAttemp` record will be deleted from the database only when the user solves the captcha. I am wondering how can this be accomplished within `axes`?

### LINKED ISSUE #224: Add option for configuring number of proxies for end user IP calculation
# HTTP header injection vulnerability in Axes

At the moment we implement a naive default of using the first IP address in `X-Forwarded-For` address list for calculating the end user IP address. In fact, this should depend on the number of proxies between the server and the client. For example, if we are running behind a nginx proxy or an [AWS Elastic Load Balancer](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html), the proxy *appends* the client address to the *end* of the list of given `X-Forwarded-For` headers. This leads to the current version of Axes having a [XFF spoofing vulnerability](https://www.whitehatsec.com/blog/top-3-proxy-issues-that-no-one-ever-told-you/).

### Vulnerability description

Clients can attempt a higher number of logins than they should by setting, for example, the X-Forwarded-For header to various different values (known as header injection or spoofing). Axes treats each different spoofed value as an unique client.

For example, let's say the proxy and client supply the following headers:

#### Case 1:  a trusted proxy appends client IP with no preset when we have one (1) known proxy in the chain

    X-Forwarded-For: client-ip-appended-by-trusted-proxy

#### Case 2: two (2) known proxies inserting first a proxy and then the client IP into the chain

    X-Forwarded-For: client-ip-appended-by-trusted-proxy, proxy-ip-appended-by-trusted-proxy

#### Case 3: malicious client injects (spoofs) a malicious X-Forwarded-For header into the request when we have one (1) trusted proxy, proxy adds actual client IP last

    X-Forwarded-For: 127.0.0.1, client-ip-appended-by-trusted-proxy

Axes treats the first IP, `127.0.0.1`, as an unique client; client can test multiple logins this way

#### Case 4: malicious client injects (spoofs) a malicious X-Forwarded-For header into the request when we have two (2) trusted proxies, proxies append client and proxy IPs last in the chain

    X-Forwarded-For: 127.0.0.1, client-ip-appended-by-trusted-proxy, proxy-ip-appended-by-trusted-proxy

Axes treats `127.0.0.1` as an unique client; client can, again, test multiple logins this way

### Current functionality

We have a bunch of different scenarios and we can have any other header configured as the client IP denominator, such as `X-Real-IP` by nginx of the default `X-Forwarded-For` which is often used with Apache, nginx, ELB, and the like.

We always just use the first IP in the defined header to calculate the client IP and disregard different configurations for number of proxies Axes can be deployed with.

### Fix

We should provide a configurable value for setting the number of proxies that would default to `0` proxies if not set; [this would be compatible with the current behaviour](https://github.com/jazzband/django-axes/blob/master/axes/decorators.py#L65).

If we could configure the IPs we could implement the following and count the IP addresses from the end of the list:

```python
from django.conf import settings
from django.core.exceptions import SuspiciousOperation

# Axes settings are configured, at the moment, with the following defaults:

BEHIND_REVERSE_PROXY = getattr(settings, 'AXES_BEHIND_REVERSE_PROXY', False)
REVERSE_PROXY_HEADER = getattr(settings, 'AXES_REVERSE_PROXY_HEADER', 'X-Forwarded-For')

# We could add the following default configuraton value:

NUM_PROXIES = getattr(settings, 'AXES_NUM_PROXIES', 0)

# IP resolution is corrected with a different IP fetching implementation

def get_ip(
        request, 
        num_proxies=NUM_PROXIES, 
        header_name=REVERSE_PROXY_HEADER,
        behind_reverse=BEHIND_REVERSE_PROXY
):
    """Calculate client IP dynamically from given request
    
    This function allows for custom header and number of proxies
    to be used when calculating end user IPs from HTTP headers.

    For example, take the following values:

        request.META['X-Forwarded-For'] = '127.0.0.1, 10.10.10.10'
        num_proxies = 1
        header_name = 'X-Forwarded-For'
        behind_reverse = True

    We can dependably calculate the end user IP from these:

        ip := '10.10.10.10'
            = ['127.0.0.1', '10.10.10.10'][-1]           
            = [addr.strip() for addr in '127.0.0.1, 10.10.10.10'].split(',')][-1]
            = [addr.strip() for addr in request.META['X-Forwarded-For'].split(',')][-1]
            = [addr.strip() for addr in request.META[header_name].split(',')][-num_proxies]
    """

    if behind_reverse:
        header_value = request.META[header_name]
        addresses = [addr.strip() for addr in header_value.split(',')]
        if addresses < num_proxies:
            raise SuspiciousOperation('Invalid request header {}: {}'.format(header_name, header_value)) 
        return addresses[-num_proxies]
    return request.META['REMOTE_ADDR']
```

### Effects for current users if a fix is upstreamed

Unaffected. 

- Fetching the `-0` IP from the IP address list is equivalent to fetching `0`, so all existing apps would be unaffected unless configured with new `AXES_NUM_PROXIES` settings. 
- This would still work with e.g. `X-Real-IP` header like `X-Real-IP: client-ip-by-nginx` and with settings set to something `AXES_NUM_PROXIES = 0` OR `AXES_NUM_PROXIES = 1` and `AXES_REVERSE_PROXY_HEADER = 'X-Real-IP'`

The functionality is implemented in configurable fashion in [Django REST Framework](http://www.django-rest-framework.org/api-guide/throttling/#how-clients-are-identified) and other software such as Werkzeug, and [Django middleware exists with XFF correction](https://github.com/codetry/xff).

aleksihakli:
This enables secure calculation of client IP value by allowing the end users to set the number of proxies they have in their current setups.

Fixes #224


camilonova:
@aleksihakli Please fix the conflicts

aleksihakli:
@camilonova rebased and updated. It seems my logging PR #226 was out of date and referenced a wrong issue number #221 but that has been fixed here. I think the captcha issue was also answered, though :)

aleksihakli:
Hey @camilonova would it be possible to merge this? It includes tests and some cleanup for the earlier PR.

camilonova:
@aleksihakli Thank you

PR #228, opened 2017-04-23
Fixes #222, and robust testing for lockout conditions

### LINKED ISSUE #222: Not blocking user only attempts.
User Only Failures is not blocking the 4th and later login attempts.
After 3 attempts the user is redirected to the lockout_template once, BUT....
If you reload the login page, UNLIMITED requests for the same username are allowed.
I can even provide the correct login credentials and authenticate the user during this "cool off time".

Commenting out the ONLY_USER_FAILURES and reloading the server works as expected when blocking by IP.
``` python
AXES_ONLY_USER_FAILURES = True
AXES_LOGIN_FAILURE_LIMIT = 3
AXES_USERNAME_FORM_FIELD = 'username'
AXES_LOCK_OUT_AT_FAILURE = True
# AXES_COOLOFF_TIME = 1  # one hour (Integer values define only hours in AXES)
AXES_COOLOFF_TIME = timedelta(minutes=15)  # 15 minutes (as timedelta object for minutes)
AXES_LOCKOUT_TEMPLATE = 'registration/_locked.html'
```

Python 2.7.12
Django==1.10.6
django-axes==2.3.2

HuntedCodes:
This PR adds a series tests to ensure robust blocking behavior under various conditions, both with and without the cache enabled, to localize diagnosis in the future. 

It also fixes issue #222 so that `AXES_ONLY_USER_FAILURES` works correctly, and a previously unknown issue with `LOCK_OUT_BY_COMBINATION_USER_AND_IP`.

Here is a truth table showing what the tests verify:

```
               User       IP           Action
             |--------------------------------
IP Only      | Same       Same         Block
(Default)    | Same       Different    Allow
             | Different  Same         Block
             | Different  Different    Allow
             |--------------------------------
User Only    | Same       Same         Block
             | Same       Different    Block
             | Different  Same         Allow
             | Different  Different    Allow
             |--------------------------------
User and IP  | Same       Same         Block
             | Same       Different    Allow
             | Different  Same         Allow
             | Different  Different    Allow
```

Another thing that this PR does is related to the tests. The default cooldown for all tests was set to 2 seconds, which caused unpredictable results in Travis CI. If a test took longer than the cooldown, the targeted behavior couldn't be accurately verified. The default cooldown for tests is now none, and the two tests that verify the cooldown behavior itself have it set to 2 seconds.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -499,23 +499,30 @@ def get_cache_key(request_or_object):
     :param  request_or_object: Request or AccessAttempt object
     :return cache-key: String, key to be used in cache system
     """
-    ua = None
-    ip = None
-
     if isinstance(request_or_object, AccessAttempt):
         ip = request_or_object.ip_address
+        un = request_or_object.username
         ua = request_or_object.user_agent
     else:
         ip = get_ip(request_or_object)
+        un = request_or_object.POST.get(USERNAME_FORM_FIELD, None)
         ua = request_or_object.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
 
-    ip = ip.encode('utf-8')
+    ip = ip.encode('utf-8') if ip else ''
+    un = un.encode('utf-8') if un else ''
+    ua = ua.encode('utf-8') if ua else ''
 
-    if ua:
-        ua = ua.encode('utf-8')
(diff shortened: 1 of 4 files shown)


camilonova:
Thank you

PR #254, opened 2017-10-12
Use user.is_authenticated method if possible

bbayles:
This PR changes `axes.decorators.check_request` such that it uses the `User.is_authenticated` method when running under Django 1.10 and above.

The callable `is_authenticated()` method was deprecated in Django 1.10 - see the notice [here](https://docs.djangoproject.com/en/1.10/releases/1.10/#using-user-is-authenticated-and-user-is-anonymous-as-methods). This PR prevents this warning from showing up:
```
RemovedInDjango20Warning: Using user.is_authenticated() and user.is_anonymous() as a method is deprecated. Remove the parentheses to use it as an attribute.
```

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -3,6 +3,7 @@ import logging
 from socket import inet_pton, AF_INET6, error
 from hashlib import md5
 
+from django import VERSION as DJANGO_VERSION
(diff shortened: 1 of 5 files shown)


camilonova:
Thank you. At the new version this is no longer needed.

PR #256, opened 2017-11-10
decorators.py: use __repr__ in get_client_str function

Bu3a3a:
Use __repr__ in get_client_str function to avoid UnicodeEncodeError in cases when username, user-agent or path_info contains non-ascii characters.


camilonova:
Thank you. We drop support for django 2.x so this wont be necessary.

PR #257, opened 2017-11-10
Fix DeprecationWarning for logger warning

richardowen:
This is currently emitting a DeprecationWarning and is deprecated in the documentation here:

https://docs.python.org/3/library/logging.html#logging.Logger.warning


camilonova:
Thank you. This was done at 1be6356 because of the changes from the current version.

PR #258, opened 2017-11-14
Switch to django-appconf

EvaSDK:
Basically a smaller changeset than the work done in django-1.11 branch.
My goal was to make the project testable via tox so I could assert if it is currently working with Django 1.11 or not and hopefully get things moving from there rather than "huge" breaking changes. This does break things if someone was crazy enough to try to patch django-axes configuration directly, but I do not think this should hold a release (say 2.4.0) for Django 1.11 support.


camilonova:
Thank you. This was already done for the new version.

PR #259, opened 2017-11-14
Allow running unittests via tox

EvaSDK:
Based on https://github.com/jazzband/django-axes/pull/258, this pull request intends to allow running unittests locally with tox so you are not forced to push and wait for Travis CI results.


camilonova:
Thank you. This was already done.

PR #262, opened 2017-11-17
Fixes #261: Global lockout possibility in 2.3.3

### LINKED ISSUE #261: Global lockout possibility in 2.3.3 when AXES_ONLY_USER_FAILURES=True
Since 2.3.3, it's possible to lockout every user from any IP. Steps to reproduce:

a) `AXES_ONLY_USER_FAILURES` or `LOCK_OUT_BY_COMBINATION_USER_AND_IP` is set to `True`
b) Caching is used (new in 2.3.3)
c) You try to login with an empty username and reach the failure limit

The reason for this lockout is two-fold:

1. When you fill in an empty username, the login form validation raises an Exception. This is seen as an incorrect login attempt by Django-Axes, of a user with username `''` (empty string).
2. The cache key created for this is basically `axes-`. Since `AXES_ONLY_USER_FAILURES=True` only the username is stored in the cache key, which is empty.

When you now go to the login page, the `is_already_locked` function is called. No username is provided since we just did a GET request on the login page. This function then matches the cache key mentioned in (2) and the page returns a locked out message. This effectively locks out everyone, from any IP.

Reverting back to 2.3.2 fixes this problem.

joeribekker:
See #261 
Happens when `AXES_ONLY_USER_FAILURES` or `LOCK_OUT_BY_COMBINATION_USER_AND_IP` is set to `True` and someone reaches the lockout limit by passing an empty username in the login form.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -433,6 +433,9 @@ def lockout_response(request):
 def is_already_locked(request):
     ip = get_ip(request)
 
+    if (AXES_ONLY_USER_FAILURES or LOCK_OUT_BY_COMBINATION_USER_AND_IP) and request.method == 'GET':
+        return False
+
     if NEVER_LOCKOUT_WHITELIST and ip_in_whitelist(ip):
         return False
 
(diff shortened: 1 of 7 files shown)


camilonova:
Thank you. I had to write them because of the big changes recently made.

Issue #266, opened 2017-11-22
Consider restoring support for Python 2.7


pope1ni:
Would you consider restoring support for Python 2.7?
Some of us are unable to upgrade to Python 3 just yet.

- Django 1.8 is the previous LTS and supported until at least April 2018.
- Django 1.10 is end-of-life December 2017.
- Django 1.11 is the current LTS and supported until at least April 2020.
- Django 2.0 drops support for Python 2.

Django 1.11 is the last version to support Python 2.7 and there is no version of django-axes that supports both of these.

jezdez:
@pope1ni Agreed, this seems premature, Python 2.7 will be most likely maintained by distros for a foreseeable future as well.

jezdez:
Closing in favor of #264.

pope1ni:
Thanks @jezdez.

Issue #264, opened 2017-11-20
Compability with older Django and Python versions


joeribekker:
Although I encourage dropping support for older versions of Python and Django, I find the requirement for Django-Axes to work **only** with Python 3.6 (according to CHANGES.txt) and Django 1.11 a bit extreme... specially since there are far more versions of Django that are still supported, as well as Python versions.

In my opinion, there should be at least a branch that supports these versions and get bugfixes (like #261). The project works fine with Python 3.5 so I don't see any reason not to add this version to Travis and support it.

jezdez:
Hey @camilonova, we're aiming at supporting older versions of Django and Python as long as they are supported by one of the Django releases. Let's re-add Python 2.7 support.

camilonova:
@joeribekker Could you help me out working on a branch adding the support? I'm aware of using some python 3 at the decorator and I'm not sure what else is needed.

Since I don't have much time to work on this I would prefer someone else to take it.

EvaSDK:
I pushed PR #258, #259 and #260 specifically to allow easier transition from 2.3 to 3 :sob: Not everyone can jump to python 3 right away ­­­— I wish I could — as a lot of Django apps are not ready for it yet.

pope1ni:
Thanks for restoring support for Python 2.7.

Sorry to be a pain, but we should probably also restore support for Python 3.4 with Django 2.0 and follow the supported versions of Python that Django does as the standard. Pull request incoming...

EvaSDK:
Thanks for the quick fixes guys. Now to get through the upgrade guide :)

PR #268, opened 2017-11-23
Test against Python 2.7.

mbaechtold:
As discussed in https://github.com/jazzband/django-axes/issues/264 and https://github.com/jazzband/django-axes/issues/266.

We need to install `mock` when testing against Python 2.7 because `mock` has been added to the standard lib only starting with Python 3.3.


camilonova:
Thank you

PR #269, opened 2017-11-23
Test against Python 3.4.

pope1ni:
As discussed in #264.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: python
 cache: pip
 python:
+- 3.4
 - 3.5
 - 3.6
 install: pip install tox-travis
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: python
 cache: pip
 python:
+- 2.7
 - 3.4
 - 3.5
 - 3.6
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 envlist =
     py{27,35,36}-django-111
```
mbaechtold: According to https://docs.djangoproject.com/en/1.11/faq/install/#what-python-version-can-i-use-with-django, Django 1.11 should run with Python 3.4 too. Could you please add it too?

camilonova: Please add here 34 as well

.travis.yml
```diff
@@ -2,6 +2,7 @@ sudo: false
 language: python
 cache: pip
 python:
+- 2.7
```
mbaechtold: Good catch, thank you!

tox.ini
pope1ni: Yes. Quite right!

pope1ni: I think you both got there at the same time!

pope1ni:
This should be all good pending the green tick...

pope1ni:
So this looks good, as the tests pass, but we haven't restored anything such as use of `six.u()` (although use of `@python_2_unicode_compatible` from `django.utils.encoding` would be better.)

It may be that we can just get away without this which will make life easier.

@camilonova: Would it be possible to release v3.0.3 with the restored support if you're happy with this?

camilonova:
Sure.

PR #274, opened 2017-11-30
Fixed logging failed attempts using a custom authenication backend.

D3X:
A quick background story:
We are using SSO with out Django site. Previous version of `axes` was working fine. I tried upgrading to 3.0.3 today and spotted an issue where it tries to create an `AccessLog` object based on `request` and `credentials['username']`. Neither of those is provided to `authenticate` by our backend, so the signal handler raised an exception.

I think adding a simple check like this should be a good enough solution. People using custom backends that don't rely on passing `request` or use differently named fields in `credentials` should implement their own log handlers if they wish to.

### DIFF
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -31,6 +31,10 @@ user_locked_out = Signal(providing_args=['request', 'username', 'ip_address'])
 def log_user_login_failed(sender, credentials, request, **kwargs):
     """ Create an AccessAttempt record if the login wasn't successful
     """
+    if request is None or 'username' not in credentials:
+        log.info('Attempt to authenticate with a custom backend failed.')
+        return
+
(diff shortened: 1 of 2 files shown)


axes/signals.py
```diff
@@ -31,6 +31,10 @@
 def log_user_login_failed(sender, credentials, request, **kwargs):
     """ Create an AccessAttempt record if the login wasn't successful
     """
+    if request is None or 'username' not in credentials:
+        log.info('Attempt to authenticate with a custom backend failed.')
```
camilonova: should we use `log.error` ?

D3X: Makes sense, updated.

camilonova:
Thank you

Issue #273, opened 2017-11-27
Remove monkey-patched LoginView methods in apps.py


JosephKiranBabu:
In apps.py, two methods of LoginView have been monkey-patched with the decorated versions from django-axes. This makes the library methods unusable as they have been replaced with their decorated versions. This is a serious problem in django applications that do not use the vanilla django LoginView and create other LoginViews by inheriting from django LoginView.
Also, now django-axes can no longer be used with django-simple-captcha since a new LoginView with CaptchaFields cannot be created.

We had to create workarounds for this problem (https://github.com/freedombox/Plinth/pull/1155/files#diff-731aef8cc92c441d1d49eed734cb4f46) to be able to use django-simple-captcha.

This monkey-patching makes a huge assumption that users of this library never create inherited LoginViews. The documentation should be updated to let the users know how to decorate their LoginViews themselves.

camilonova:
I had the same feeling when doing the changes but I couldn't find a better way to do it.

If you have a better implementation please send a PR.

Thanks.

JGONGSQ:
I have the same issue. I am using the Django 1.8.16 and its not working.

D3X:
@camilonova, @JosephKiranBabu  

I can think of two possible solutions for this issue:

1. Add a settings like `AXES_MONKEYPATCH_DJANGO_AUTH` and only patch the methods if it's `True`. 

    Pros:
    + Super-easy to do (a single `if` in `apps.py` should be enough)
    + Does not break BC

    Cons:
    - It's still monkeypatching, so it's rather dirty

2. Provide an `AxesLoginView` that inherits from `LoginView` and overrides said methods. That would require people to change the way they use `axes` now, but I like it better, because it's explicit and there's no monkeypatching. The two overridden methods could even be provided by a mixin, like this:
    ```python
    class AxesLoginMixin(object):
        @method_decorator(axes_dispatch)
        def dispatch(self, request, *args, **kwargs):
            return super(AxesLoginMixin, self).dispatch(request, *args, **kwargs)

        @meethod_decorator(axes_form_invalid)
        def form_invalid(self, form):
            return super(AxesLoginMixin, self).form_invalid(form)

    class AxesLoginView(AxesLoginMixin, LoginView):
        pass
    ```

    That way, users using custom `LoginView`s could simply add a mixin to their classes to get the desired behaviour.

    Pros:
    + No monkeypatching
    + Allows for more customisation in the future
[...]

PR #277, opened 2017-12-08
Do not monkeypatch LoginView methods.

D3X:
Related to #273 and freedombox/Plinth#1138

Up for a discussion. If we decide to go with this, I'll add tests to cover both the legacy setting and the new view.


axes/apps.py
```diff
@@ -1,16 +1,28 @@
...
 class AppConfig(apps.AppConfig):
     name = 'axes'
 
     def ready(self):
-        from django.contrib.auth.views import LoginView
-        from django.utils.decorators import method_decorator
+        from . import signals    # NOQA: Load the signals module to bind the receivers
+
+        if settings.USE_LEGACY_LOGIN_VIEW_PATCH:
+            log.warning('Automatic patching of LoginView using AXES_USE_LEGACY_LOGIN_VIEW_PATCH is deprecated. '
```
aleksihakli: You should use DeprecationWarning for this.

https://docs.python.org/3.6/library/exceptions.html#DeprecationWarning

aleksihakli:
Looks good aside from using DeprecationWarnings for deprecating features which can be easily tested and are catchable with the `python` `-W` flag. I think implementing pull requests and fixing & testing the infinite redirect bug are essential to this PR, so those should be included in the initial work.

camilonova:
@D3X Could you please take a look again?

D3X:
@camilonova Sure. I kinda forgot about this PR. Will take a look over the weekend.

axes/apps.py
```diff
@@ -1,16 +1,36 @@
...
+from .conf import settings
+
+
+log = logging.getLogger(settings.AXES_LOGGER)
+
+
+def patch_login_view():
+    warnings.warn(
+        'Automatic patching of LoginView using AXES_USE_LEGACY_LOGIN_VIEW_PATCH is deprecated. '
+        'use axes.views.AxesLoginView instead, or add axes.views.AxesLoginMixin to your custom login view.',
```
aleksihakli: Capitalization `use` -> `Use`.

axes/apps.py
```diff
@@ -1,16 +1,36 @@
...
+        DeprecationWarning
+    )
+    from django.contrib.auth.views import LoginView
```
aleksihakli: These could be imported on the top of the module, no need to separate the imports module and function level here imo.

axes/views.py
```diff
@@ -0,0 +1,24 @@
+from __future__ import absolute_import
+
+from django.contrib.auth.views import LoginView
+
+from axes.attempts import is_already_locked
```
aleksihakli: This should preferably be `from .attempts import is_already_locked`.

D3X:
@aleksihakli Yeah, I noticed the tests failing. I tried fixing them yesterday, but realised it requires changing a bit more than I initially thought.

The `AdminSite.login` view uses `LoginView` directly, so we'd have to somehow make it use the new `AxesLoginView`. To avoid patching it, we probably have to provide an `AxesAdminSite` class. I don't know how to make it work with other 3rd party apps that register their models using `admin.site.register`, where `admin.site` is the `AdminSite` instance provided by Django.

aleksihakli:
Huh, [looks like](https://github.com/django/django/blob/master/django/contrib/admin/sites.py#L374) [so](https://github.com/django/django/blob/master/django/contrib/admin/sites.py#L349), and the loading of the login and logout views can not be parametrized either via Django other than by subclassing the admin site.

I think the best way to approach this is to both

1. keep the current behaviour and monkey patching of the login views, and
2. offer the new mixins and a custom admin site with documentation.

The first option is what most people will probably use and the second offers more flexibility and customization options. 

Some users will probably benefit from the custom views and admins greatly, and most users will never use them.

The problem with implementing a custom admin is that it increases the amount of code and complexity of this package. Do we want and need that? Should we just document the approaches on implementing the custom views with our mixins and views so people can integrate `axes` into their 3rd party packages and custom admins and implement custom authentication themselves?

In my opinion the custom admin site and mixins are neat and relate closely to the core idea of `axes`, but the growing complexity is a big minus. 

I suggest we should fix the tests and implement the mixins and `AxesLoginView` here but leave the admin site customization up for another PR once this work is merged into upstream.

axes/test_settings.py
```diff
@@ -75,3 +75,4 @@
 LOGIN_REDIRECT_URL = '/admin/'
 
 AXES_FAILURE_LIMIT = 10
+AXES_USE_LEGACY_LOGIN_VIEW_PATCH = False
```
aleksihakli: This should be `True` to allow old test logic to pass.

axes/tests/test_apps.py
```diff
@@ -0,0 +1,25 @@
+import warnings
+
+from django.contrib.auth.views import LoginView
+from django.test import TestCase
+
+from axes.apps import patch_login_view
+
+
+class PatchLoginViewTest(TestCase):
```
aleksihakli: Use [override_settings()](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.override_settings) to use the newer login view logic.

axes/tests/test_views.py
```diff
@@ -0,0 +1,47 @@
+from unittest.mock import Mock, patch
+
+from django.contrib.auth.models import User
+from django.test import TestCase, RequestFactory
+from django.utils.six.moves import http_client
+
+from axes.views import AxesLoginView
+
+
+class AxesLoginViewTest(TestCase):
```
aleksihakli: Use [override_settings()](https://docs.djangoproject.com/en/2.0/topics/testing/tools/#django.test.override_settings) to use the newer login view logic.

axes/conf.py
```diff
@@ -55,3 +55,5 @@ class MyAppConf(AppConf):
     IP_WHITELIST = None
 
     IP_BLACKLIST = None
+
+    USE_LEGACY_LOGIN_VIEW_PATCH = True
```
aleksihakli: This should probably be just `PATCH_LOGIN_VIEW` which would translate to `AXES_PATCH_LOGIN_VIEW` if we are unsure about the deprecation at the moment.

markddavidoff:
@D3X @aleksihakli maybe take a look at my approach using an authentication backend [here](https://github.com/jazzband/django-axes/pull/315)?

PR #278, opened 2017-12-10
Fix #272 which was erroneously closed.

### LINKED ISSUE #272: Cannot import watch_login
The command below is not working when l installed the latest package. I checked the decorator and it doesn't exist

eschercode:
Make usage cases (using django.contrib.auth's signals) explicit, and also detail how version 3 works, so that previous version 2 users who expect the `watch_login` decorator aren't confused.
Issue #272 was opened due to incorrect documentation. Unfortunately, the relevant documentation file was deleted rather than updated, and the issue was closed. This commit is intended to fix the missing
documentation.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -2,9 +2,24 @@
 
 Usage
 =====
+``django-axes`` listens to signals from ``django.contrib.auth.signals`` to
+log access attempts:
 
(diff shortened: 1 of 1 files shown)


camilonova:
Thank you

PR #282, opened 2017-12-15
Added support for custom username field

kakulukia:
This adds the ability to also use other fields like email as the username field.

### DIFF
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.contrib.auth import get_user_model
 from django.contrib.auth.signals import user_logged_in
 from django.contrib.auth.signals import user_logged_out
 from django.contrib.auth.signals import user_login_failed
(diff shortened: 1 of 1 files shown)


axes/signals.py
```diff
@@ -27,16 +28,26 @@
 user_locked_out = Signal(providing_args=['request', 'username', 'ip_address'])
 
 
+def get_username_field():
+    try:
+        username_field = get_user_model().USERNAME_FIELD
```
camilonova: This works since django 1.7 I think we can safely asume it will work, so we don't need the try and except block

axes/signals.py
```diff
@@ -27,16 +28,26 @@
...
+
+    return username_field
+
+
 @receiver(user_login_failed)
 def log_user_login_failed(sender, credentials, request, **kwargs):
     """ Create an AccessAttempt record if the login wasn't successful
     """
-    if request is None or 'username' not in credentials:
+    username_field = get_username_field()
```
camilonova: change to `username_field = get_user_model().USERNAME_FIELD`

camilonova:
Thank you

camilonova:
I took a better look at it and we had a setting when this happens. Please use the `AXES_USERNAME_FORM_FIELD` if needed.

kakulukia:
But its not used in the code i altered .. and the setting is obsolete just mirroring the users username field, isn't it?

kakulukia:
or to not make too many changes AXES_USERNAME_FORM_FIELD could at least default to User.USERNAME_FIELD

Issue #288, opened 2018-01-07
axes_reset ip  Does not restore access to address


razyr:
When I run

```
manage.py axes_reset ip 123.123.123.123
```
Axes reports "1 attempts removed".

When I check, the related accessattempt object has been removed from the database as expected.  However, the IP remains blocked.   It seems to stay blocked until I restart the application docker container, at which point the address is able to connect again.

If I use Django admin to remove the related accessattempt object, the IP is immediately allowed to connect again.

This seems to be consistent and repeatable behaviour.

aleksihakli:
Access uses the [`axes.utils.reset`](https://github.com/jazzband/django-axes/blob/da0c4b429a2065c562419b94342e0accdc64e104/axes/utils.py#L53) function for resetting the login attempts from the `manage` command and resets the cache with [`axes.attempts.get_cache_key`](https://github.com/jazzband/django-axes/blob/da0c4b429a2065c562419b94342e0accdc64e104/axes/attempts.py#L52) whereas the admin site just deletes the `AccessAttempt` objects related to the IP and the [`axes.signals.delete_cache_after_delete`](https://github.com/jazzband/django-axes/blob/da0c4b429a2065c562419b94342e0accdc64e104/axes/signals.py#L171) signal handler handles the deletion of cache keys.

@camilonova I think we should remove the extra deletion logic from the `axes.utils.reset` function since it already exists in `axes.signals.delete_cache_after_delete` signal handler, and investigate deeper into this potential bug.

There could be a flaw in the `AccessAttempt` reset / deletion code and it seems like we have duplicate deletion logic which fires on both deletion types, actually. There is a slight difference in the way the deletion is handled.

@razyr could you post your Docker container setup here as well, including your `Dockerfile` and `docker-compose` setups? Which cache backend and database are you using? Is this reproducible on dummy cache, in-memory cache, and Memcached / Redis cache backends?

razyr:
This is a low traffic application, so we're just using the default in-memory cache.  The reported problem disappears with dummy (no) caching.

Our database is Postgres 9.4 as shown below in docker-compose.yml.   

Our dev Docker configuration is quite minimal with simply an application container and a database container.  We have a more sophisticated production.yml which adds an NGINX front end, but the issue reported here is consistent in both our dev and production environments with default caching.

Dockerfile
```
FROM python:3.6.3
ENV PYTHONUNBUFFERED 1
RUN mkdir /code
WORKDIR /code
ADD . /code/
[...]

PR #292, opened 2018-01-14
Refactor AccessAttempt reset utility

aleksihakli:
Remove cache key deletion because cache resets are already done in the [delete_cache_after_delete](https://github.com/jazzband/django-axes/blob/918f040dd71540130d790b5ee8fa1dcf70c73409/axes/signals.py#L173) signal handler after AccessAttempt object deletion.

Hopefully affects #288 positively.

---

Could you help verify this @razyr? Just use this line in your `requirements.txt` instead of `django-axes`: 

    https://github.com/jazzband/django-axes/archive/44cb07814e27ec01b86e56ee5ecd92a66fb6cc8f.zip

Or install it with `pip` on top of the current `django-axes` package:

    pip install --upgrade https://github.com/jazzband/django-axes/archive/44cb07814e27ec01b86e56ee5ecd92a66fb6cc8f.zip

### DIFF
--- a/axes/utils.py
+++ b/axes/utils.py
@@ -54,7 +54,6 @@ def reset(ip=None, username=None):
     """Reset records that match ip or username, and
     return the count of removed attempts.
     """
-    count = 0
 
     attempts = AccessAttempt.objects.all()
     if ip:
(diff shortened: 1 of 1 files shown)


razyr:
Unfortunately, this does not fix #288.  Same behaviour as the released version 4.0.1.

I tested it on both our dev and production systems.

In both cases, I confirmed that the new code was in place by manually inspecting utils.py in the container under /usr/local/lib/python3.6/site-packages/axes.

Deleting an entry through Django Admin does immediately restore access.

Deleting an entry through 'manage.py axes_reset ip' does not restore access until the container/application is restarted to clear the cache.

aleksihakli:
Thanks for testing @razyr. It should be a cache mechanism or bug somewhere then, will have to look harder to locate it.

@camilonova I think this is good for merging anyway and cleans up the current code base, could you merge if you think it's good?

aleksihakli:
Continuing in #293.

PR #293, opened 2018-01-14
Refactor caching and related documentation

aleksihakli:
Related to #288, improve the default access attempt handling and add documentation on the known problems with `LocMemCache`.


camilonova:
Thank you

PR #295, opened 2018-01-14
Fix inet_pton can't be imported on windows (update patch)

hsiaoyi0504:
Replace #291. To solve #204 and #208.

PR #296, opened 2018-01-15
Fix inet_pton can't be imported on windows

aleksihakli:
Revise #295 syntax and pip qualifier usage to use the same methods in the `setup.py` file as in the `axes/utils.py` file.

I ran the `tox` test set manually on Windows, seems to work as expected on Python 2.7. Running without `win_inet_pton` imported produces the error described.


camilonova:
Thank you

PR #303, opened 2018-02-09
Add default cache backend checks to avoid incorrect axes behaviour

aleksihakli:
As the incorrect LocMemCache caching has seemingly affected multiple people using the default caching settings I think we should raise an ImproperlyConfigured error if axes is used with it as the default cache backend.

The `default` cache backend should always be set, so I think this simple check mitigates the most common errors with AXES for people who do not invest time in reading the documentation of the plugins they are using.

@camilonova could you please review this patchset? It does introduce backwards incompatible changes but on the other hand fixes clearly erronous use of the `LocMemCache` for many users.

### DIFF
--- a/axes/apps.py
+++ b/axes/apps.py
@@ -5,6 +5,12 @@ class AppConfig(apps.AppConfig):
     name = 'axes'
 
     def ready(self):
+        from django.conf import settings
+        from django.core.exceptions import ImproperlyConfigured
+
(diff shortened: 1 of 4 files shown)


camilonova:
Beautiful. Thanks

aleksihakli:
@charleswhchan I agree, but that will require a lot more work to implement and especially test all the branches on caching. I think this was a good starting point for the work. Please feel free to contribute on the caching configuration work :)

PR #305, opened 2018-02-11
Fix #299

charleswhchan:
This patch fixes #299. Since LocMemCache instance is created per process, the failure count might end up being incorrect if there are multiple Django processes. This fix checks if LocMemCache is used as the caching backend and bypass the cache and rely on database record instead.

Test (before applying fix):
```
AXES: BEGIN LOG
AXES: Using django-axes 4.0.2
AXES: blocking by IP only.
Creating test database for alias 'default'...
System check identified no issues (0 silenced).
AXES: New login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Creating access record.
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.2.2.2', user-agent: '<unknown>', path: '/admin/login/'}. Count = 2 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Count = 2 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.2.2.2', user-agent: '<unknown>', path: '/admin/login/'}. Count = 3 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Count = 3 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.2.2.2', user-agent: '<unknown>', path: '/admin/login/'}. Count = 4 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Count = 4 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.2.2.2', user-agent: '<unknown>', path: '/admin/login/'}. Count = 5 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Count = 5 of 10
AXES: Repeated login failure by {user: 'valid-user-1', ip: '10.1.1.1', user-agent: '<unknown>', path: '/admin/login/'}. Count = 6 of 10
F
======================================================================
FAIL: test_lockout_by_user_blocks_when_same_user_diff_ip_using_locmemcache (axes.tests.test_access_attempt_config.AccessAttemptConfigTest)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

Issue #302, opened 2018-02-09
Make caching optional


aleksihakli:
Contuing the discussion in #299.

@t1user it would be OK to make the caching implementation optional and that could, in fact, make the code cleaner. 

Using `LocMemCache` should raise a warning or error on startup to prevent invalid usage.

aleksihakli:
At the moment the caching can be disabled by using the `DummyCache` backend. The caching on the module could also be made optional with simple flags, though.

charleswhchan:
I would like to see a flag be added to axes config to control cache behaviour 😃

PR #306, opened 2018-02-11
Fix #299 #302

### LINKED ISSUE #299: Possible bug in signals.py?
https://github.com/jazzband/django-axes/blob/master/axes/signals.py#L59

Should the following:
```
        for attempt in attempts:
            failures = max(failures, attempt.failures_since_start)
```

be modified to:
```
        for attempt in attempts:
            failures += attempt.failures_since_start
```

Given that settings.AXES_ONLY_USER_FAILURES maybe return multiple attempts from the same user (different path or data). Wouldn't it make sense for all failed attempts to be summed together?  🤔 💭 

Thanks!

charleswhchan:
Added a new option AXES_USE_CACHE to workaround the issue with LocMemCache.
Enables axes to bypass the cache.

Perhaps this is a better solution than #305? Looking forward to the feedback. 😃

### DIFF
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -47,3 +47,5 @@ class MyAppConf(AppConf):
     IP_WHITELIST = None
 
     IP_BLACKLIST = None
+
+    USE_CACHE = True
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -52,7 +52,8 @@ def log_user_login_failed(sender, credentials, request, **kwargs):
     cache_timeout = get_cache_timeout()
 
     failures_cached = cache.get(cache_hash_key)
-    if failures_cached is not None:
+    if failures_cached is not None and \
+        settings.AXES_USE_CACHE:
         failures = failures_cached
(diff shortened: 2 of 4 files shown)


aleksihakli:
Need some time to review this work. There might be more places where caching is used and those should be revised as well. Thank you for looking into this so quickly!

charleswhchan:
Sure. Looking forward to the feedback 👍

aleksihakli:
I'm not sure how necessary this changeset is #307. It should check for LocMemCache being used in the setup phase and fail configuration of `axes` altogether. A more important thing would be to add the option for disabling caches altogether.

charleswhchan:
Thanks @aleksihakli. I am happy to see a better solution.

I did examine #307 in detail, but it seems like it provides an option to allow the axes cache to be set independently from the rest of Django -- ie. to DummyCache. In this case, it will effectively resolve the issue with #209 and #302 (sort of).

PR #307, opened 2018-02-12
Merlin

JWvDronkelaar:
Django-axes is incompatible with locmemcache which is the default cache backend in our project (for now). To be able to use django-axes regardless we added a setting AXES_CACHE which allows us to set a different cache just for axes. This seems to have solved our issue for now, perhaps it is useful for other users as well.

I noticed an ongoing discussion about a flag to disable caching altogether. In my opinion this would be a nice option to have as well.

### DIFF
--- a/axes/utils.py
+++ b/axes/utils.py
@@ -1,30 +1,18 @@
-import logging
-
 from platform import python_version
 from sys import platform
-
 if python_version() < '3.4' and platform == 'win32':
     import win_inet_pton
 from socket import inet_pton, AF_INET6, error
 
-from django.core.cache import cache, caches, InvalidCacheBackendError
+from django.core.cache import cache, caches
 from django.utils import six
 
 from axes.conf import settings
 from axes.models import AccessAttempt
 
 
-log = logging.getLogger(settings.AXES_LOGGER)
-
-
 def get_axes_cache():
-    axes_cache_name = settings.AXES_CACHE
-    if axes_cache_name:
-        axes_cache = caches[axes_cache_name]
-    else:
-        axes_cache = cache
-
-    return axes_cache
+    return caches[getattr(settings, 'AXES_CACHE', 'default')]
 
 
 def query2str(items, max_length=1024):
--- a/axes/utils.py
+++ b/axes/utils.py
@@ -1,15 +1,30 @@
+import logging
+
 from platform import python_version
 from sys import platform
+
 if python_version() < '3.4' and platform == 'win32':
     import win_inet_pton
 from socket import inet_pton, AF_INET6, error
 
-from django.core.cache import cache
+from django.core.cache import cache, caches, InvalidCacheBackendError
 from django.utils import six
 
-from axes.attempts import get_cache_key
 from axes.conf import settings
 from axes.models import AccessAttempt
+
+
+log = logging.getLogger(settings.AXES_LOGGER)
+
+
+def get_axes_cache():
+    axes_cache_name = settings.AXES_CACHE
+    if axes_cache_name:
+        axes_cache = caches[axes_cache_name]
+    else:
+        axes_cache = cache
+
+    return axes_cache
 
 
 def query2str(items, max_length=1024):
--- a/merlin.txt
+++ b/merlin.txt
@@ -1,0 +1,1 @@
+Make cache backend that django-axes uses configurable through a setting.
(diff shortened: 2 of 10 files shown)


axes/utils.py
```diff
@@ -1,17 +1,32 @@
...
-from axes.attempts import get_cache_key
 from axes.conf import settings
 from axes.models import AccessAttempt
 
 
+log = logging.getLogger(settings.AXES_LOGGER)
+
+
+def get_axes_cache():
+    axes_cache_name = settings.AXES_CACHE
```
aleksihakli: Could just be written as

```python
return caches[getattr(settings, 'AXES_CACHE', 'default')]
```

merlin.txt
```diff
@@ -0,0 +1 @@
+Make cache backend that django-axes uses configurable through a setting.
```
aleksihakli: Is this intended to be included in the PR?

axes/utils.py
```diff
@@ -1,17 +1,32 @@
...
+
 from platform import python_version
 from sys import platform
+
 if python_version() < '3.4' and platform == 'win32':
     import win_inet_pton
 from socket import inet_pton, AF_INET6, error
 
-from django.core.cache import cache
+from django.core.cache import cache, caches, InvalidCacheBackendError
```
aleksihakli: InvalidCacheBackendError is not used anywhere.

axes/utils.py
```diff
@@ -1,17 +1,32 @@
+import logging
+
```
aleksihakli: Unnecessary line changes.

axes/utils.py
```diff
@@ -1,17 +1,32 @@
...
 from platform import python_version
 from sys import platform
+
```
aleksihakli: Please include style changes in a separate PR.

axes/utils.py
JWvDronkelaar: Agreed, will change.

merlin.txt
JWvDronkelaar: No it is not, I will remove it.

axes/utils.py
JWvDronkelaar: Yes it is a remnant from an error I wanted to catch. I will remove it.

axes/utils.py
JWvDronkelaar: I will revert it.

axes/utils.py
JWvDronkelaar: Sure no problem.

JWvDronkelaar:
Hi Aleksi! In short I agree with everything you suggested. I will make the changes you suggested. I don't know if I have the time to do much additional work on it on top of those changes.

That said it would be great if either cache choice or the disabling of cache altogether could become a part of this official library. It would make the requirements for our software a lot simpler/reliable (instead of having to install from a fork at a specific commit).

aleksihakli:
I think this changeset looks nice now. What do you think @camilonova?

aleksihakli:
@JWvDronkelaar this line should be changed to use the `AXES_CACHE` as well:

https://github.com/jazzband/django-axes/blob/master/axes/apps.py#L11

Then the checks for LocMemCache should function correctly.

aleksihakli:
Also, could you add documentation to docs/ prior to merging about the `AXES_CACHE` setting and an example of its usage? Using DummyCache as the `AXES_CACHE` also effectively disables caching with this changeset.

JWvDronkelaar:
Hello @aleksihakli . I have made the requested changes to the docs. The ready check now also takes the AXES_CACHE setting into account. For this I had to merge the Merlin branch with the updated master branch and I noticed the latest commit to master does not show here. Should I have commited it to the Merlin branch that I merged? I am new to forks and their workflow. If there is something I need to do to make this work for you please let me know.

JWvDronkelaar:
@aleksihakli I have asked my colleague for advice and merged the master into the merlin branch instead which seems to give a better result. I hope this is the correct process, if not I'd love to hear from you on how to fix it.

camilonova:
@aleksihakli looks good to me.

@JWvDronkelaar Great work pal! I would recommend to squash your commits into a single one for clarity purposes.

aleksihakli:
Yeah, squash the commits and with `git rebase` and I think this is good for merging :)

aleksihakli:
Hey, this has been now rebased and merged in #309 and you are the author of the cache work @JWvDronkelaar. I'm closing this PR since the content is already in master. Thank you!

PR #309, opened 2018-02-17
Add AXES_CACHE setting

### LINKED ISSUE #299: Possible bug in signals.py?

### LINKED ISSUE #302: Make caching optional

### LINKED ISSUE #304: Question: How is cache configured in test_access_attempts_config.py?
While crafting a test/fix for #299, discover something unusual about the test case:

In [test_access_attempts_config.py](https://github.com/jazzband/django-axes/blob/master/axes/tests/test_access_attempt_config.py):
```
    def test_lockout_by_ip_allows_when_same_user_diff_ip_without_cache(self):
        # User 1 is locked out from IP 1.
        self._lockout_user1_from_ip1()

        # User 1 can still login from IP 2.
        response = self._login(
            self.USER_1,
            self.VALID_PASSWORD,
            ip_addr=self.IP_2
        )
        self.assertEqual(response.status_code, self.ALLOWED)
```

```
    def test_lockout_by_ip_allows_when_same_user_diff_ip_using_cache(self):
        # User 1 is locked out from IP 1.
        self._lockout_user1_from_ip1()

        # User 1 can still login from IP 2.
        response = self._login(
            self.USER_1,
            self.VALID_PASSWORD,
            ip_addr=self.IP_2
        )
        self.assertEqual(response.status_code, self.ALLOWED)
```

Other than the name of the test, the test cases appears to be identical. How does the test initialize|reset|enable|disable cache? I am sure there is a perfectly reasonable explanation and I am simply missing something obvious. 😅 

Thanks!

aleksihakli:
Rebased and cleaned up #307 on top of master.

This should also fix #305 and fix #306 with the LocMemCache checks and close #304, close #302, and close #299.

### DIFF
--- a/axes/apps.py
+++ b/axes/apps.py
@@ -8,8 +8,10 @@ class AppConfig(apps.AppConfig):
         from django.conf import settings
         from django.core.exceptions import ImproperlyConfigured
 
-        if settings.CACHES['default']['BACKEND'] == 'django.core.cache.backends.locmem.LocMemCache':
-            raise ImproperlyConfigured('django-axes does not work properly with LocMemCache as the default cache backend')
+        if settings.CACHES[getattr(settings, 'AXES_CACHE', 'default')]['BACKEND'] == \
+                'django.core.cache.backends.locmem.LocMemCache':
+            raise ImproperlyConfigured(
+                'django-axes does not work properly with LocMemCache as the default cache backend')
 
         from django.contrib.auth.views import LoginView
         from django.utils.decorators import method_decorator
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -2,13 +2,13 @@ from datetime import timedelta
 from hashlib import md5
 
 from django.contrib.auth import get_user_model
-from django.core.cache import cache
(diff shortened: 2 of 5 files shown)


aleksihakli:
Hey @camilonova, I rebased the `merlin` branch and work on top of the current master. Could you review and merge this if it looks good?

camilonova:
Thank you!

PR #315, opened 2018-03-13
use an authentication backend to check if user is locked out.

markddavidoff:
@camilonova First Draft for input.
### USAGE
settings.py
```
AUTHENTICATION_BACKENDS = [
    'axes.middleware.DjangoAxesAuthBackend',
    'django.contrib.auth.backends.ModelBackend',
]
```
login view call example
```
user = User.objects.get(email=form.cleaned_data['email'])
response_context = {}
user = authenticate(
    request=request,
    username=user.username,
    password=form.cleaned_data['password'],
    response_context=response_context
)
...
if 'error' not in context:
    # if don't have a more specific error, use a generic one.
    context['error'] = 'Incorrect Login Information'
render(request, 'log_in.html', context)
```

### DIFF
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -21,6 +21,9 @@ class AxesModelBackend(ModelBackend):
         :return: Nothing, but will update return_context with lockout message if user is locked out.
         """
 
+        if request is None:
+            raise ValueError('DjangoAxesModelBackend requires calls to authenticate to pass `request`')
+
         if is_already_locked(request):
             # locked out, don't try to authenticate, just update return_context and return
             # Its a bit weird to pass a context and expect a response value but its nice to get a "why" back.
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -1,0 +1,32 @@
+from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import PermissionDenied
+
+from axes.attempts import is_already_locked
+from axes.utils import get_lockout_message
+
+
+class AxesModelBackend(ModelBackend):
+
+    def authenticate(self, request, username=None, password=None, **kwargs):
+        """
+        Add django-axes handling and add allow adding errors directly to a passed return_context.
+        Will never actually authenticate a user, just blocks locked out uses so don't use this as your only back end.
+        :param request: see ModelBackend.authenticate
+        :param username: see ModelBackend.authenticate
+        :param password: see ModelBackend.authenticate
+        :keyword response_context: context dict that will be returned/used in the response template.
+            NOTE: will overwrite 'error' field in dict
+        :param kwargs: see ModelBackend.authenticate
+        :raises PermissionDenied: if user is already locked out.
+        :return: Nothing, but will update return_context with lockout message if user is locked out.
+        """
+
+        if is_already_locked(request):
+            # locked out, don't try to authenticate, just update return_context and return
+            # Its a bit weird to pass a context and expect a response value but its nice to get a "why" back.
+            error_msg = get_lockout_message()
+            response_context = kwargs.get('response_context', {})
+            response_context['error'] = error_msg
+            raise PermissionDenied(error_msg)
+
+        # No-op
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -7,6 +7,13 @@ from axes.utils import get_lockout_message
 
 class AxesModelBackend(ModelBackend):
 
+    class RequestParameterRequired(Exception):
+        msg = 'DjangoAxesModelBackend requires calls to authenticate to pass `request`'
+
+        def __init__(self):
+            super(AxesModelBackend.RequestParameterRequired, self).__init__(
+                AxesModelBackend.RequestParameterRequired.msg)
+
     def authenticate(self, request, username=None, password=None, **kwargs):
         """
         Add django-axes handling and add allow adding errors directly to a passed return_context.
@@ -22,7 +29,7 @@ class AxesModelBackend(ModelBackend):
         """
 
         if request is None:
-            raise ValueError('DjangoAxesModelBackend requires calls to authenticate to pass `request`')
+            raise AxesModelBackend.RequestParameterRequired()
 
         if is_already_locked(request):
             # locked out, don't try to authenticate, just update return_context and return
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -47,3 +47,9 @@ class MyAppConf(AppConf):
     IP_WHITELIST = None
 
     IP_BLACKLIST = None
+
+    # message to show when locked out and have cooloff enabled
+    AXES_COOLOFF_MESSAGE = 'Account locked: too many login attempts. Please try again later'
+
+    # message to show when locked out and have cooloff disabled
+    AXES_PERMALOCK_MESSAGE = 'Account locked: too many login attempts. Contact an admin to unlock your account.'
--- a/axes/utils.py
+++ b/axes/utils.py
@@ -92,3 +92,10 @@ def iso8601(timestamp):
         if value]
     )
     return 'P' + date + ('T' + time if time else '')
+
+
+def get_lockout_message():
+    if settings.AXES_COOLOFF_TIME:
+        return settings.AXES_COOLOFF_MESSAGE
+    else:
+        return settings.AXES_PERMALOCK_MESSAGE
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -3,7 +3,9 @@
 Configuration
 =============
 
-Just add `axes` to your ``INSTALLED_APPS``::
+3 Simple Steps!
+
+- add `axes` to your ``INSTALLED_APPS``::
 
     INSTALLED_APPS = (
         'django.contrib.admin',
@@ -16,7 +18,26 @@ Just add `axes` to your ``INSTALLED_APPS``::
         ...
     )
 
-Remember to run ``python manage.py migrate`` to sync the database.
+- add Axes to the top of ``AUTHENTICATION_BACKENDS``::
+
+    AUTHENTICATION_BACKENDS = [
+        'axes.middleware.DjangoAxesAuthBackend',
+        ...
+        'django.contrib.auth.backends.ModelBackend',
+        ...
+    ]
+
+- run ``python manage.py migrate`` to sync the database.
+
+Things to you might need to change in your code, especially if you get a ``AxesModelBackend.RequestParameterRequired``:
+
+- make sure any calls to ``django.contrib.auth.authenticate`` pass the request.
+
+- make sure any auth libraries you use that call the authentication middleware stack pass request. Notably Django Rest
+  Framework (DRF) ``BasicAuthentication`` does not pass request. `Here is an example workaround for DRF`_.
+
+.. _Here is an example workaround for DRF: http://www.python.org
+
 
 Known configuration problems
 ----------------------------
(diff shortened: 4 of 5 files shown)


axes/middleware.py
```diff
@@ -0,0 +1,29 @@
+from axes.attempts import is_already_locked
+from axes.utils import get_lockout_message
+from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import PermissionDenied
+
+class DjangoAxesAuthBackend(ModelBackend):
```
markddavidoff: Name suggestions?

axes/middleware.py
```diff
@@ -0,0 +1,29 @@
...
+        :param kwargs: see ModelBackend.authenticate
+        :raises PermissionDenied: if user is already locked out.
+        :return: Nothing, but will update return_context with lockout message if user is locked out.
+        """
+
+        if is_already_locked(request):
+            # locked out, don't try to authenticate, just update return_context and return
+            error_msg = get_lockout_message()
+            return_context = kwargs.get('response_context', {})
+            return_context['error'] = error_msg
```
markddavidoff: I know this is a bit weird, but I like the option to get messages back from the backend.

camilonova:
Looks good so far, please keep going.

axes/middleware.py
```diff
@@ -0,0 +1,30 @@
...
+        :param kwargs: see ModelBackend.authenticate
+        :raises PermissionDenied: if user is already locked out.
+        :return: Nothing, but will update return_context with lockout message if user is locked out.
+        """
+
+        if is_already_locked(request):
+            # locked out, don't try to authenticate, just update return_context and return
+            error_msg = get_lockout_message()
+            response_context = kwargs.get('response_context', {})
+            response_context['error'] = error_msg
```
aleksihakli: Does this mutate kwargs contents?

axes/conf.py
```diff
@@ -47,3 +47,9 @@ class MyAppConf(AppConf):
     IP_WHITELIST = None
 
     IP_BLACKLIST = None
+
+    # message to show when locked out and have cooloff enabled
+    AXES_COOLOFF_MESSAGE = 'Account locked: too many login attempts. Please try again later'
+
+    # message to show when locked out and have cooloff disabled
+    AXES_PERMALOCK_MESSAGE= 'Account locked: too many login attempts. Contact an admin to unlock your account.'
```
aleksihakli: Missing a newline.

axes/utils.py
```diff
@@ -92,3 +92,10 @@ def iso8601(timestamp):
         if value]
     )
     return 'P' + date + ('T' + time if time else '')
+
+
+def get_lockout_message():
+    if settings.AXES_COOLOFF_TIME:
+        return settings.AXES_COOLOFF_MESSAGE
+    else:
+        return settings.AXES_PERMALOCK_MESSAGE
```
aleksihakli: Missing a newline.

axes/middleware.py
markddavidoff: It expects a dict called 'response_context' to be passed as a kwarg and mutates that if it is present.

 I agree its a bit weird to pass a context and expect a value set but its nice to get a "why" back and I couldn't figure out a better way to do this that still complies with the django auth backend structure.

axes/conf.py
markddavidoff: also missing a space on the `=`

axes/middleware.py
```diff
@@ -0,0 +1,30 @@
+from axes.attempts import is_already_locked
+from axes.utils import get_lockout_message
+from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import PermissionDenied
+
+
+class DjangoAxesAuthBackend(ModelBackend):
```
aleksihakli: Maybe `AxesModelBackend` would be a neater name?

axes/middleware.py
```diff
@@ -0,0 +1,30 @@
+from axes.attempts import is_already_locked
+from axes.utils import get_lockout_message
+from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import PermissionDenied
```
aleksihakli: A more traditional import order would be to import Python core packages first, then Django packages and then custom packages. Something like

    from django import ...

    from axes import ...

axes/middleware.py
markddavidoff: sure! I copy pasted a bunch of this from my subclass implementation and rushed to redact some of the naming and messages

axes/middleware.py
```diff
@@ -0,0 +1,30 @@
+from axes.attempts import is_already_locked
```
aleksihakli: This file is not middleware - it should be `backends.py` so the import for this module would be something like `axes.backends.AxesModelBackend`.

aleksihakli:
Looks good, the naming is a bit conflicting as this patchset implements an authentication backend instead of middleware (though it's easy to see it as middlewareish in functionality).

The approach, however, is _very_ nice. An authentication backend denying logins is MUCH more neater and universal than monkey patching view dispatchers which is the current functionality.

If you could clean this up slightly and implement unit tests for the backend it would be nice to get this merged :+1:

camilonova:
@markddavidoff can we have this soon? Looks amazing.

axes/backends.py
```diff
@@ -21,6 +21,9 @@ def authenticate(self, request, username=None, password=None, **kwargs):
         :return: Nothing, but will update return_context with lockout message if user is locked out.
         """
 
+        if request is None:
```
markddavidoff: @camilonova @aleksihakli addressed pr concerns, but i have another query as to handle when request is None.

Django Rest Framework (which is very widely used) does not set `request` when passing it to authenticate, so things like @authentication_classes((BasicAuthentication,)) will error out. 

options:

- Error out like this
- no-op if no request and log a warning

aleksihakli: I think this behaviour is fine for the time being. If someone wants to use this custom backend in their project they need to make sure it can function correctly.

aleksihakli:
Looks good to me! I think this is ready for merging. I'll let @camilonova check and merge this :)

axes/backends.py
```diff
@@ -7,6 +7,13 @@
 
 class AxesModelBackend(ModelBackend):
 
+    class RequestParameterRequired(Exception):
```
aleksihakli: This exception should really live in `axes.exceptions` as per good naming conventions. You just create that `exceptions.py` file. There is no need to put this inside the `AxesModelBackend` class, nesting classes just makes this harder to read and find.

docs/configuration.rst
```diff
@@ -16,7 +18,26 @@ Just add `axes` to your ``INSTALLED_APPS``::
...
+- run ``python manage.py migrate`` to sync the database.
+
+Things to you might need to change in your code, especially if you get a ``AxesModelBackend.RequestParameterRequired``:
+
+- make sure any calls to ``django.contrib.auth.authenticate`` pass the request.
+
+- make sure any auth libraries you use that call the authentication middleware stack pass request. Notably Django Rest
+  Framework (DRF) ``BasicAuthentication`` does not pass request. `Here is an example workaround for DRF`_.
+
+.. _Here is an example workaround for DRF: http://www.python.org
```
aleksihakli: This doesn't seem like the right link.

docs/configuration.rst
```diff
@@ -16,7 +18,26 @@ Just add `axes` to your ``INSTALLED_APPS``::
...
+    AUTHENTICATION_BACKENDS = [
+        'axes.middleware.DjangoAxesAuthBackend',
+        ...
+        'django.contrib.auth.backends.ModelBackend',
+        ...
+    ]
+
+- run ``python manage.py migrate`` to sync the database.
+
+Things to you might need to change in your code, especially if you get a ``AxesModelBackend.RequestParameterRequired``:
```
aleksihakli: This can also be just `axes.exceptions.RequestParameterRequired`.

axes/backends.py
```diff
@@ -0,0 +1,42 @@
+from django.contrib.auth.backends import ModelBackend
+from django.core.exceptions import PermissionDenied
+
+from axes.attempts import is_already_locked
+from axes.utils import get_lockout_message
+
+
+class AxesModelBackend(ModelBackend):
+
+    class RequestParameterRequired(Exception):
```
aleksihakli: This should be moved into `axes.exceptions` file (you can just create that) as per good Django package conventions. Exceptions are easier to find when they are in standard locations.

axes/backends.py
```diff
@@ -0,0 +1,42 @@
...
+
+        if is_already_locked(request):
+            # locked out, don't try to authenticate, just update return_context and return
+            # Its a bit weird to pass a context and expect a response value but its nice to get a "why" back.
+            error_msg = get_lockout_message()
+            response_context = kwargs.get('response_context', {})
+            response_context['error'] = error_msg
+            raise PermissionDenied(error_msg)
+
+        # No-op
```
aleksihakli: We could check for the request authentication status in the model backend and lock out if there are duplicate requests.

aleksihakli:
I guess the big question is if we want to do the access checking and locking out in the model backend?

axes/backends.py
markddavidoff: Can you please clarify what you mean by 'check for the request authentication status'?

docs/configuration.rst
markddavidoff: Woops. Will fix.

axes/backends.py
markddavidoff: I strongly disagree eith moving the exception and using suspicious operation, although im always open to naming changes. 

Moving indicates it is a failure on the module level, which i dont think applies here. The convention, to me, suggests a "Misconfigured" or "BadParameter" parent exc in exceptions.py that can be subclassed here, but i think the subclass should live here as it is tied to the backend not the module. 

Is this a suspicious operation? I think its more programmer error?

markddavidoff:
@aleksihakli, you mean as the default recommended way?

axes/backends.py
markddavidoff: But maybe im thinking app as part of a big project wise and not standalone library wise

aleksihakli:
@markddavidoff yeah! I think the best place for all that checking is the `AxesModelBackend` introduced here. That way we could catch all authentication requests and check their authentication status neatly in one place instead of monkey patching any views or affecting any other code.

markddavidoff:
@aleksihakli I'm still confused what you mean, i believe this does all the checking the monkeypatched view does already. Am i missing other logic?

axes/backends.py
aleksihakli: Never mind, I think this is already handled by our signal handler - backend combination.

axes/backends.py
aleksihakli: OK, I think it's valid to have this here as well. Would be nice to have some tests for this, but I think we can implement those later.

aleksihakli:
@markddavidoff if the [Django authentication signals](https://docs.djangoproject.com/en/2.0/ref/contrib/auth/#module-django.contrib.auth.signals) which have handlers defined in `axes/signals.py` fire correctly, then `axes` gets information about user login attempts and logouts. Then this backend should work nicely. 

I wondered if we should attempt to do access attempt logging in the backend level, but I suppose that isn't necessary in the case all signals work correctly. We should be able to remove the monkey patching with this change from `axes` altogether if everything works nicely with this new backend.

I'll go ahead and merge this changeset, let's refine it further in a separate PR if necessary :)

aleksihakli:
It seems that `authenticate` does not indeed return a 403 but just skips the authentication:

https://github.com/django/django/blob/master/django/contrib/auth/__init__.py#L62

So we will either have to modify the tests that check response code on an attempted login, implement a custom middleware that returns the 403 code, or continue monkey patching stuff.

markddavidoff:
This is meant to be used in conjunction with another middleware that
actually authenicates like in the example i added to the docs. We can use
modelbackend in the tests. Any suggestions to make that more clear in code
and docs.

PR #316, opened 2018-03-13
use username from credentials and not from request.POST to query attempts.

markddavidoff:
any reason to pull them from the request.POST ?


markddavidoff:
hmm, this needs some more work, closing for now.

Issue #286, opened 2018-01-02
Potential security issues with `get_ip()` changes in v4


pope1ni:
I'm a little concerned by the change to use `ipware.ip.get_ip()` in v4.

1. There is no documentation to suggest that `django-ipware` may need to be configured.
2. There is now no documentation on the potential attack vector if the proxy doesn't set `X-Forwarded-For` (or whatever the first header used in `IPWARE_META_PRECEDENCE_ORDER` is) explicitly, c.f. [`USE_X_FORWARDED_HOST`](https://docs.djangoproject.com/en/2.0/ref/settings/#std:setting-USE_X_FORWARDED_HOST)
3. ~~`django-ipware` says to use `get_ip()` if the server is **not** accessible on the Internet, see [here](https://github.com/un33k/django-ipware#how-to-use).~~ (Although looking at the code the distinction between `get_ip()` and `get_real_ip()`, the latter doesn't count private IP addresses as "real". What they really mean here is use `get_ip()` if there are also clients that may connect from an internal private network otherwise use `get_real_ip()`.)
4. Previously `django-axes` allowed configuring the maximum number of proxies in case of spoofed entries. This is no longer possible. `django-ipware` provides `get_trusted_ip()` but it isn't particularly useful in it's current form.

It would probably be wise to recommend setting `IPWARE_META_PRECEDENCE_ORDER` to `['HTTP_X_FORWARDED_FOR', 'REMOTE_ADDR']` or at least indicate that this was the original behaviour. Checking all of these headers by default feels risky.

Forgive me, but this feels rushed and ill thought through. It was working fine for me and issue #275 and pull request #280 fail to even give any evidence of a problem. Every comment in #275 is regarding `AttributeError` when looking up `META` on `NoneType` - `django-ipware` is still using `request.META`. This feels like a push to remove code in favour of a dependency for minor gain.

aleksihakli:
Hey, `django-ipware` indeed seems to be somewhat prone to proxy header injection and doesn't have the number of known proxies configurable. I previously contributed #225 which fixed IP address resolving by calculating the IP based on the number of proxies in the proxy chains, but using `django-ipware` scrubs this. Configuration for number of proxies is useful in e.g. AWS where there might be requests coming in directly to the web server reverse proxy whereas we would want request to flow through load balancers and into the web server; the configuration for the number fixes this and filters out unwanted requests with invalid headers.

[...]

PR #322, opened 2018-04-10
Add django-ipware documentation

aleksihakli:
Relates to #286


aleksihakli:
Already added in another commit.

PR #325, opened 2018-04-11
Add proxy and client IP configuration settings and documentation

### LINKED ISSUE #286: Potential security issues with `get_ip()` changes in v4

aleksihakli:
Fixes #286

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -4,18 +4,16 @@ from hashlib import md5
 from django.contrib.auth import get_user_model
 from django.utils import timezone
 
-from ipware.ip2 import get_client_ip
-
 from axes.conf import settings
 from axes.models import AccessAttempt
-from axes.utils import get_axes_cache
+from axes.utils import get_axes_cache, get_client_ip
 
 
 def _query_user_attempts(request):
     """Returns access attempt record if it exists.
     Otherwise return None.
     """
-    ip, _ = get_client_ip(request)
+    ip = get_client_ip(request)
     username = request.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
 
     if settings.AXES_ONLY_USER_FAILURES:
@@ -60,7 +58,7 @@ def get_cache_key(request_or_obj):
         un = request_or_obj.username
         ua = request_or_obj.user_agent
     else:
-        ip, _ = get_client_ip(request_or_obj)
+        ip = get_client_ip(request_or_obj)
         un = request_or_obj.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
         ua = request_or_obj.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
 
@@ -176,7 +174,7 @@ def is_user_lockable(request):
 
 
 def is_already_locked(request):
-    ip, _ = get_client_ip(request)
+    ip = get_client_ip(request)
 
     if (
         settings.AXES_ONLY_USER_FAILURES or
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -53,3 +53,23 @@ class MyAppConf(AppConf):
 
     # message to show when locked out and have cooloff disabled
     PERMALOCK_MESSAGE = 'Account locked: too many login attempts. Contact an admin to unlock your account.'
+
+    # if your deployment is using reverse proxies, set this value to 'left-most' or 'right-most' per your configuration
+    PROXY_ORDER = 'left-most'
+
+    # if your deployment is using reverse proxies, set this value to the number of proxies in front of Django
+    PROXY_COUNT = None
+
+    # if your deployment is using reverse proxies, set to your trusted proxy IP addresses prefixes if needed
+    PROXY_TRUSTED_IPS = None
+
+    # set to the names of request.META attributes that should be checked for the IP address of the client
+    # if your deployment is using reverse proxies, ensure that the header attributes are securely set by the proxy
+    # ensure that the client can not spoof the headers by setting them and sending them through the proxy
+    META_PRECEDENCE_ORDER = getattr(
+        settings, 'AXES_META_PRECEDENCE_ORDER', getattr(
+            settings, 'IPWARE_META_PRECEDENCE_ORDER', (
+                'REMOTE_ADDR',
+            )
+        )
+    )
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -3,7 +3,7 @@
 Configuration
 =============
 
-Add `axes` to your ``INSTALLED_APPS``::
+Add ``axes`` to your ``INSTALLED_APPS``::
 
     INSTALLED_APPS = (
         'django.contrib.admin',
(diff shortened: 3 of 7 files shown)


aleksihakli:
@pope1ni @un33k @camilonova @SunilMohanAdapa @EvaSDK @vdboor since this has been a source of much discussion, please check out this PR that relies on the recent work enhancement added by @un33k and adds configurable proxy count and cleaner request.META lookup configuration.

I think this should satisfy the needs for strict header checking while maintaining backwards compatibility in `ipware.ip2.get_client_ip`.

pope1ni:
It would also probably be worth exposing the `proxy_trusted_ips` parameter of `get_client_ip()` to Django Axes which would satisfy the point raised in https://github.com/jazzband/django-axes/issues/286#issuecomment-380438315 about trusted proxies.

axes/conf.py
```diff
@@ -53,3 +53,11 @@ class MyAppConf(AppConf):
...
     # message to show when locked out and have cooloff disabled
     PERMALOCK_MESSAGE = 'Account locked: too many login attempts. Contact an admin to unlock your account.'
+
+    # set to the amount of proxies you are using if you have reverse proxies configured
+    IPWARE_PROXY_COUNT = None
+
+    # set to a trusted header value if your reverse proxy is configured to set it to the HTTP request
+    IPWARE_REQUEST_HEADER_ORDER = (
+        'REMOTE_ADDR',
+    )
```
vdboor: This leaves a project with two settings to configure:
* `IPWARE_REQUEST_HEADER_ORDER`
* `AXES_IPWARE_REQUEST_HEADER_ORDER`

I'd rather see secure defaults in django-ipware, because now axes is "forking" the settings to protect itself, whil leaving other packages to their fate. Before we know it, we have a `ALLAUTH_IPWARE_REQUEST_HEADER_ORDER` and `RATELIMIT_IPWARE_REQUEST_HEADER_ORDER`. If you'd still take this rate, at least use a configured `settings.IPWARE_REQUEST_HEADER_ORDER` as fallback in case it's configured. As this is such security point, this should use a really simple solution, and not a hairy set of settings to configure.

pope1ni: I agree, but I've struggled enough to get the point across as it is and this is a workable solution.

Django Axes use to have it's own settings - albeit named differently - for these things when it had it's own implementation, so it isn't the worst thing.

This also does have the minor benefit that Axes can have it's own isolated configuration - someone may want to configure IPware to use a different header to look up a different IP for some other reason...

axes/attempts.py
```diff
@@ -60,7 +58,7 @@ def get_cache_key(request_or_obj):
         un = request_or_obj.username
         ua = request_or_obj.user_agent
     else:
-        ip, _ = get_client_ip(request_or_obj)
+        ip = get_client_ip(request_or_obj)
```
un33k: It would be best if IP is cached on the first request and used on subsequent calls.  Generating cache key should not be more costly than getting the IP.

axes/conf.py
un33k: It might be better to give advanced users ability to overwrite, and fallback on defaults for the rest.

```python
from django.conf import settings

AXES_META_PRECEDENCE_ORDER = getattr(settings,
    'AXES_META_PRECEDENCE_ORDER', getattr(settings,
        'IPWARE_META_PRECEDENCE_ORDER', (
             'REMOTE_ADDR',
              # ...
        )
     )
  )
}

# AXES_META_PRECEDENCE_ORDER is used so people who don't care about ipware don't have to know about it.
# IPWARE_META_PRECEDENCE_ORDER is used so people who already use ipware don't have to worry about it.
# Ipware can take all the params via kwargs, and it can be use directly. No need to worry about mixed settings.
```

docs/configuration.rst
```diff
@@ -27,11 +27,7 @@ Add ``axes.backends.AxesModelBackend`` to the top of ``AUTHENTICATION_BACKENDS``
 
 Run ``python manage.py migrate`` to sync the database.
 
-Configure `django-ipware <https://github.com/un33k/django-ipware/>`_ to your liking. Pay close attention to the  `IPWARE_META_PRECEDENCE_ORDER <https://github.com/un33k/django-ipware#precedence-order>`_ setting. Please note that this configuration is required for functional security in your project. A good starting point for a project running without a reverse proxy could be::
-
-    IPWARE_META_PRECEDENCE_ORDER = (
-        'REMOTE_ADDR',
-    )
+Configure `AXES_IPWARE_PROXY_COUNT` and `AXES_IPWARE_REQUEST_HEADER_ORDER` as per `django-ipware <https://github.com/un33k/django-ipware/>`_ instructions if you are running behind a reverse proxy. Pay close attention to the  `AXES_IPWARE_REQUEST_HEADER_ORDER <https://github.com/un33k/django-ipware#precedence-order>`_ setting. Please note that this configuration is very important for functional security in your project.
```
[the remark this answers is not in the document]
un33k: All new variables and settings are better to start with AXES_, and should not have any IPWARE in them. Less confusion as well as the fact that ipware is a reserved name of a different package.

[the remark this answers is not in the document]
un33k: Also, setup.py should point to **django-ipware>=2.0.2**

axes/conf.py
SunilMohanAdapa: Without something like this, our application will be broken by this change. Our application already has a fix in place by setting `IPWARE_META_PRECEDENCE_ORDER=('HTTP_X_FORWARDED_FOR',),`.

axes/attempts.py
aleksihakli: I added caching to a custom request object attribute `axes_request_client_ip` in the `get_client_ip` utility function since the current use of cache simply gets the number of requests from the `axes` cache, and does not account for client IP address. Storing these values into the Django cache would require some more refactoring which is unessential for this change.

docs/configuration.rst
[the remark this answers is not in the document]
aleksihakli: Fixed the version pinning.

The variables loaded from `conf.py` are actually prefixed with `AXES_` by the `django-appconf` package, the configuration class is a subclass of `appconf.Appconf`.

PR #329, opened 2018-04-18
Backport unicode string literals for Python 2.7

### LINKED ISSUE #328: Encoding error when usernames contain non-ascii characters
Reason: <https://github.com/jazzband/django-axes/blob/master/axes/utils.py#L38-L39>

Quickfix when using axes: Changing the Setting `AXES_VERBOSE = False`

Adding a `u`-prefix to the details string might fix this for py2 versions while keeping py3.3+ compatibility.

aleksihakli:
Fixes #328

### DIFF
--- a/axes/__init__.py
+++ b/axes/__init__.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 __version__ = '4.2.0'
 
 default_app_config = 'axes.apps.AppConfig'
--- a/axes/admin.py
+++ b/axes/admin.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from django.contrib import admin
 
 from axes.models import AccessLog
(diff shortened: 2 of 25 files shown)


aleksihakli:
@camilonova I think we should support Python 2.7 if we support a Django version that supports it. Python 2.7 will be deprecated in 1.5 years so I don't think we should explicitly drop support from the code base before that. The wait for Python 3 only is not that long now in mid-2018, anyway.

There are a lot of projects using Python 2.7 and if we can support those in our codebase with few simple __future__ and backport imports we should IMO do that.

This is just my philosophy. I think that libraries that have large user bases should support different language versions if it isn't a big stretch, and in this specific case it isn't.

toabi:
I'm thankful for this quick solution, I wish all my projects were py3 already…

PR #331, opened 2018-04-20
Misc lint fixes

EvaSDK:
Fix some issues as suggested by pylint.
The one that annoyed me most being logging. We use sentry to collect log messages and analyze errors in our stack and use non-lazzy logging breaks automatic grouping of events/messages. Also current code forces rendering at message emission while it could not be needed at all according to logging configuration.


aleksihakli:
Thank you for the very good patches and cleanup :+1:

PR #332, opened 2018-04-20
Fix call to undefined variable ua in _get

EvaSDK:
Fixes access to undefined variable in `_query_user_attempts` and
respects documentation of the settings.

I wanted to go further into simplifying the code of `_query_user_attempts` but then I realized that it is not immediately clear to me what to do in all combinations of:
* `ONLY_USER_FAILURES`
* `LOCK_OUT_BY_COMBINATION_USER_AND_IP`
* `USE_USER_AGENT`

Log messages hint that only IP and username can be considered at any time, but then when does User-Agent come into play ? Wouldn't it be clearer to have these settings expressed either as a list or as cumulative boolean options, like LOCK_BY_IP, LOCK_BY_USERNAME, LOCK_BY_USER_AGENT ?

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -41,7 +41,7 @@ def _query_user_attempts(request):
         else:
             params['ip_address'] = ip
 
-        if settings.AXES_USE_USER_AGENT:
+        if settings.AXES_USE_USER_AGENT and not settings.AXES_ONLY_USER_FAILURES:
             params['user_agent'] = ua
 
         attempts = AccessAttempt.objects.filter(**params)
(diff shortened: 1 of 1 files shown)


aleksihakli:
Yeah, the combinatory logic of these is a bit of a hassle in the codebase. I'd like to refactor some bits of that query fetching and caching, too, but delving into the different cases will require some time and focus.

EvaSDK:
If you have some hints about what you would like to see, maybe I could work on it.

aleksihakli:
Had the chance to read through the source and this fix does seem to be in place. 

Further, I think we should refactor and simplify the calculation of the filter parameters in the code base, since multiple places currently use the same kind of attribute fetching and conditional logic:

- https://github.com/jazzband/django-axes/blob/master/axes/utils.py#L43
- https://github.com/jazzband/django-axes/blob/master/axes/attempts.py#L21
- https://github.com/jazzband/django-axes/blob/master/axes/attempts.py#L36
- https://github.com/jazzband/django-axes/blob/master/axes/attempts.py#L71

PR #335, opened 2018-04-21
Version 4.3.1

aleksihakli:
I added linting and prospector checks to the CI pipeline and tox environment and cleaned up some legacy from the code base. I won't be contributing much before summer or autumn after this, but I hope this makes maintaining the project a bit easier :)

It seems I forgot to put up release 4.3.0 up in the `axes.version`, but that shouldn't be much of an issue since I just baked a small release for that yesterday. This version is better for PyPI anyway.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,19 @@
 Changes
 =======
 
+4.3.1 (2018-04-21)
+------------------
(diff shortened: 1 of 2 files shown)


camilonova:
Thank you

PR #339, opened 2018-05-09
Handle unicode in usernames or user-agents

Remiz:
When a username or a user-agent contains a non-ascii character, logs may fail.

### DIFF
--- a/axes/utils.py
+++ b/axes/utils.py
@@ -38,18 +38,18 @@ def get_client_str(username, ip_address, user_agent=None, path_info=None):
     if settings.AXES_VERBOSE:
         if isinstance(path_info, tuple):
             path_info = path_info[0]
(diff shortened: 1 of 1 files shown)


aleksihakli:
That module already uses `from __future__ import unicode_literals` so those strings should be unicode by default.

Are you sure the problem is with the strings themselves and not some variable that is being encoded as unicode? Which version of axes did you test this on? Are you sure it was the latest one? What was the stack trace? What were the relevant variable types and values for the error? 

You can install axes on your own project with the `pip install -e django-axes` flag to make it editable for testing and debugging purposes.

Remiz:
My bad, I realized I was using an older version of the plugin that is missing this unicode_literals... 
Thanks for the reply, I'll close this PR.

aleksihakli:
It was added pretty recently :)

PR #341, opened 2018-05-16
django rest fix

svenauhagen:
Hi,

the function is_user_lockable is looping for authentication requests that fail from the django rest framework.

request.user is calling authenticate in the DRF and this, in turn if it fails, is generating the user signal which calls is_user_lockable again.

The patch here does work, I am not sure if it is the best way to solve the problem.

Please let me know.

Best
Sven

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -149,8 +149,18 @@ def is_user_lockable(request):
     If so, then return the value to see if this user is special
     and doesn't get their account locked out
     """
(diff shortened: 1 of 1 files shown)


camilonova:
I don't like it, seems too specific for DRF.

`request.user` should return the user instance, nothing else should happen behind scenes. Please think about it, maybe you'll find a better approach.

Issue #318, opened 2018-03-28
Lockout by username and IP does not work with custom username


jaadus:
`AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP` does not work if the username was modified before being sent to `authenticate`.

1. `signals.py` receives the `user_login_failed` signal, and correctly fetches the modified `username` from `credentials`.
( `username = credentials[settings.AXES_USERNAME_FORM_FIELD]` )
1. It then calls `get_user_attempts` with *only* the `request` as arguments.
1. Which leads to `_query_user_attempts` (in `attempts.py`) to *incorrectly* fetch `username` from `request` (returning the unmodified value).
( `username = request.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)` )
1. 0 attempts are found because the `username` in the database is the one from `credentials`, which leads to the creation of a new attempt row with every failed login, instead of incrementing the tries and eventually locking out the user.

`username` should be passed through such that it properly queries attempts.

BasAmmerlaan:
> `username` should be passed through such that it properly queries attempts.


This would also fix another problem I ran into recently, being that axes does not work with a prefixed AuthenticationForm.
The POST data contains `prefix-username`, but `credentials` received from the signal contains just `username`.
Passing the username (as set by `AXES_USERNAME_FORM_FIELD`) from the signal all the way to `_query_user_attempts` would seem like a nice fix for both these issues.

camilonova:
PRs are welcome anytime

jaadus:
@camilonova I opened #343 as my proposed solution for this problem.

Passing `username` through the stack as I initially proposed won't cut it, because of the decorator on `dispatch`, which hasn't gone through any form cleaning and can't infer how username is _supposed_ to look when it reaches authentication.

pckapps:
What if an extra setting is added to indicate what is lookup field to use? Something like AXES_USERNAME_CREDENTIAL_FIELD. If this setting is not provided it will default to AXES_USERNAME_FORM_FIELD. Should also add a warning if the field is not in the credential dictionary. I can submit a pull request if this approach will be accepted?

jaadus:
This isn't for the situation when the actual username comes in via a different field in the post body. (That situation is discussed somewhat [in the docs](https://django-axes.readthedocs.io/en/latest/usage.html#integration-with-django-allauth).)

[...]

PR #343, opened 2018-05-22
Optional callable in settings to clean username

jaadus:
Related issue: #318

Add a setting to supply a callable that can return a correct username given a request object.
This can modify or clean it in case the supplied username in the post body does not match what goes into credentials. More about the problem this is solving in the issue.

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 
 from axes.conf import settings
 from axes.models import AccessAttempt
-from axes.utils import get_axes_cache, get_client_ip
+from axes.utils import get_axes_cache, get_client_ip, get_client_username
 
 
 def _query_user_attempts(request):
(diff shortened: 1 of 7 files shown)


jorrit-wehelp:
Was looking into doing something similar, you beat me to it.
Only think I would suggest doing different is having the setting have a default value instead of None, pointing to a function that does the backward compatible AXES_USERNAME_FORM_FIELD getter.
This make it always go through the same path and the default implementation serves as an example to be overriden by users.

jaadus:
In that case, I'd still have to go through another method to handle the `None` case as I do currently, in the util `get_client_username`, to protect against clients that override the setting explicitly to equal `None`, or add another check in `AppConfig.ready` before starting, which I would really like to avoid.

So your proposed change would just add a method that does the exact same thing as the default case in `utils.py` line 75 and set that method as the initial settings value instead of `None`. I can do that, but I don't really see any added value.

camilonova:
Looks great. Thank you.
Please add some documentation on how this feature work.

camilonova:
Awesome. Thanks!

jd-cape:
Thanks for the merge!

What's the process for getting a 4.4 version out with these changes?

camilonova:
I'll cut a release ASAP

PR #357, opened 2018-09-28
Wording change as DRF made some changes

markddavidoff:


markddavidoff:
@wodCZ is this what you had in mind?

wodCZ:
does not blame current DRF revision and keeps documentation informative and useful for some use-cases.
👍 in my opinion

PR #359, opened 2018-09-28
for consistency get username from request in log_user_login_failed

markddavidoff:
for consistency use` get_client_username` in `log_user_login_failed` instead of credentials. IMHO we should be using credentials everywhere but since we're not, at least be consistent.

I would love to use credentials to get username everywhere instead, but as we use request everywhere else, `log_user_login_failed` should use it as well


camilonova:
Thank you

PR #365, opened 2018-10-24
Bug: Add missing migrations

markddavidoff:
I also added a test to make sure this doesn't happen again.

Also if 4.4.1 is released without this, people will get migration issues.


markddavidoff:
@aleksihakli @camilonova

markddavidoff:
Its been a hot sec, I'm merging this.

aleksihakli:
Great, good job on the regression test as well.

jsharpe:
Looks like the version of 4.4.1 that got pushed to pypi didn't include this patch so its missing migrations. Can a 4.4.2 release be pushed to pypi to resolve this please?

Issue #352, opened 2018-08-23
Log in attempts are not consecutive


tadamra:
I am using django-axes 4.1.0. I set  The number of failed login attempts to 5 attempts to block the user. I noticed that the login attempts are not consecutive. If the user does 4 failed login attempts followed by 1 successful attempts and then 1 failing attempt, he will be blocked all the same. 

After reviewing possible configurations for django-axes I noticed that there is no possible config I could use to enforce the behaviour I need. Is there a way to force this required behaviour? If not I would regard this as an issue, because it is more  intuitive to force the attempts to be consecutive or at least make this option configurable.

wodCZ:
Could be implemented by listening to `user_logged_in` signal and calling `reset(ip=ip)` there.

Keep in mind that this could cause security issues. Imagine that the user sits in a cafe with an attacker at the same wifi. User successfully logs in, and grants the attacker new attempts.

sagar270690:
> """
Manages incoming connections from all communication channels and routes them to the proper channel managers
"""
Can use `reset(username=user.username)` instead.

markddavidoff:
If you are only locking on IP, this could indeed give the attacker more attempts. If you are locking on more criteria we can reset on successful login as that would only give the attacker at most (2 x AXES_FAILURE_LIMIT) -1 attempts. Should be re-using teh _query_user_attempts logic in reset anyway IMHO...

tadamra:
Thanks to everyone who participated in this issue. I will close it now.

tadamra:
The issue is still there after the feature has been merged (I guess only now it could be labeled as a bug since the functionality has been implemented). I set AXES_RESET_ON_SUCCESS to True after updating the version, still the required result is not attained.

By checking the table "axes_accessattempts"  in the database I can see that the number of attempets is not reset to zero after the successful login. Actually at a successful attempt the user has to be removed from the table completely, not only set its attempts to 0, but of course it is not. 

Please contact me if anymore information is required to solve the bug.

arjenzijlstra:
I believe this is due to the fact that the code adding the `reset_on_success` setting is not included in the most recent release.

tadamra:
I noticed that is was added to the documentation on https://django-axes.readthedocs.io/en/latest/configuration.html

This means that it should be functional and it is added to the current version.

Anyway I will retest it in the next release, after verifying weather the feature is added or not.

arjenzijlstra:
[...]

Issue #363, opened 2018-10-18
axes_reset doesn't clear the cache.


markddavidoff:
I'll fix this if I get to it, but in case anyone else has time, go for it. I'd also appreciate input here on the design question below.

### The Issue
In `utils.reset` we don't clear the cache, so even if it is cleared,  lockouts persist until cache expires.
We don't want to reset the whole cache if we're clearing doe just one username, but I'm not sure on the best approach

### Approach 1
Add cache key revocation in an `AccessAttempt` `post_delete` signal. This would require closer coupling the cache key logic and the `AccessAttempt`, which I'm fine with since the cache is used to retrieve AccessAttempts and the decoupling is what caused the issue. The con is that since signals are async, the management command will return before the full cleanup is done which may be misleading to the caller as there will be a delay. We could attach another signal and wait, but that's kinda gross.

### Approach 2
Keep the cache logic separate from `AccessAttempt` and do a looped delete instead of bulk delete in `reset`, for an ip or username specific reset, deleting the cache keys as we loop. the filtered data shouldn't be too big so It shouldn't be too much of a performance hit.

markddavidoff:
After typing it out I'm for Approach 2.  The separation of logic here seems bug prone, but if we only delete in reset, we should be ok.

arjenzijlstra:
Correct me if I'm wrong, but I believe Approach 1 is already present in the code. From `axes/signals.py`:

```
@receiver(post_delete, sender=AccessAttempt)
def delete_cache_after_delete(instance, **kwargs):  # pylint: disable=unused-argument
    cache_hash_key = get_cache_key(instance)
    get_axes_cache().delete(cache_hash_key)
```

markddavidoff:
Oh! I didn't see that you are correct, I'll have to dig into why this isn't working on my end.

PR #366, opened 2018-10-26
Added functionality to reset failed attempts on successful login.

arjenzijlstra:
Adds functionality to reset the number of failed attempts on a successful login. The functionality can be enabled by setting `AXES_RESET_ON_SUCCESS` to true. The reset function makes use of `_query_user_attempts` to get the attempts to delete.

Also discussed in #352.

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -130,6 +130,13 @@ def get_user_attempts(request):
     return attempts
 
 
+def reset_user_attempts(request):
+    attempts = _query_user_attempts(request)
+    count, _ = attempts.delete()
+
+    return count
+
+
 def ip_in_whitelist(ip):
     if not settings.AXES_IP_WHITELIST:
         return False
--- a/axes/models.py
+++ b/axes/models.py
@@ -76,6 +76,7 @@ class AccessAttempt(CommonAccess):
         verbose_name = _('access attempt')
         verbose_name_plural = _('access attempts')
 
+
 class AccessLog(CommonAccess):
     logout_time = models.DateTimeField(
         _('Logout Time'),
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -16,6 +16,7 @@ from axes.attempts import get_cache_timeout
 from axes.attempts import get_user_attempts
 from axes.attempts import is_user_lockable
 from axes.attempts import ip_in_whitelist
+from axes.attempts import reset_user_attempts
 from axes.models import AccessLog, AccessAttempt
 from axes.utils import get_client_str
 from axes.utils import query2str
@@ -144,6 +145,13 @@ def log_user_logged_in(sender, request, user, **kwargs):  # pylint: disable=unus
             http_accept=http_accept,
             path_info=path_info,
             trusted=True,
+        )
+
+    if settings.AXES_RESET_ON_SUCCESS:
+        count = reset_user_attempts(request)
+        log.info(
+            'AXES: Deleted %d failed login attempts by %s.', count,
+            get_client_str(username, ip_address, user_agent, path_info)
         )
 
 
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -150,7 +150,8 @@ def log_user_logged_in(sender, request, user, **kwargs):  # pylint: disable=unus
     if settings.AXES_RESET_ON_SUCCESS:
         count = reset_user_attempts(request)
         log.info(
-            'AXES: Deleted %d failed login attempts by %s.', count,
+            'AXES: Deleted %d failed login attempts by %s.',
+            count,
             get_client_str(username, ip_address, user_agent, path_info)
         )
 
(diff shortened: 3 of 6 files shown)


axes/attempts.py
```diff
@@ -130,6 +130,13 @@ def get_user_attempts(request):
     return attempts
 
 
+def reset_user_attempts(request):
+    attempts = _query_user_attempts(request)
+    count, _ = attempts.delete()
+
```
markddavidoff: also reset cached attempts. same issue as https://github.com/jazzband/django-axes/issues/363

arjenzijlstra: I will add this today or tomorrow.

markddavidoff: disregard, I guess we are already doing this.

arjenzijlstra: I noticed, let me know if there is anything else to improve.

axes/models.py
```diff
@@ -76,6 +76,7 @@ class Meta(object):
         verbose_name = _('access attempt')
         verbose_name_plural = _('access attempts')
 
+
```
aleksihakli: Just stating that stylistic changes should not be mixed with functional changes; makes tracking regressions and bisecting harder.

axes/signals.py
```diff
@@ -146,6 +147,13 @@ def log_user_logged_in(sender, request, user, **kwargs):  # pylint: disable=unus
             trusted=True,
         )
 
+    if settings.AXES_RESET_ON_SUCCESS:
+        count = reset_user_attempts(request)
+        log.info(
+            'AXES: Deleted %d failed login attempts by %s.', count,
```
aleksihakli: Preferably put the `count` argument on its own line.

axes/tests/test_access_attempt.py
```diff
@@ -394,3 +394,49 @@ def test_custom_authentication_backend(self):
...
+        # But we should get one now
+        response = self._login()
+        self.assertContains(response, self.LOCKED_MESSAGE, status_code=403)
+
+    @override_settings(AXES_RESET_ON_SUCCESS=True)
+    def test_reset_on_success(self):
+        """Tests that the failure attempts resets after one successful
+        attempt when using the corresponding setting.
+        """
+        print('RESET = TRUE')
```
aleksihakli: Obsolete print statement.

axes/tests/test_access_attempt.py
```diff
@@ -394,3 +394,49 @@ def test_custom_authentication_backend(self):
         request.user = self.user
         authenticate(request=request, foo='bar')
         self.assertEqual(AccessLog.objects.all().count(), 0)
+
+    # by default, AXES_RESET_ON_SUCCESS = False
+    def test_reset_on_success_default(self):
+        """Tests that the failure attempts does not reset after one successful
+        attempt by default.
+        """
+        print('RESET = FALSE')
```
aleksihakli: Obsolete print statement.

axes/tests/test_access_attempt.py
```diff
@@ -394,3 +394,49 @@ def test_custom_authentication_backend(self):
...
+        # test until one try before the limit
+        for _ in range(1, settings.AXES_FAILURE_LIMIT):
```
aleksihakli: Just `range(settings.AXES_FAILURE_LIMIT - 1)` would be cleaner to read IMO, but that's just a preference.

axes/tests/test_access_attempt.py
```diff
@@ -394,3 +394,49 @@ def test_custom_authentication_backend(self):
         request.user = self.user
         authenticate(request=request, foo='bar')
         self.assertEqual(AccessLog.objects.all().count(), 0)
+
+    # by default, AXES_RESET_ON_SUCCESS = False
+    def test_reset_on_success_default(self):
```
aleksihakli: This is exactly the same implementation as the function above this. Refactor these into a single common function that is called from both test cases and not prefixed `test` so it doesn't get executed automatically. Something like `_assert_resets_on_success` or similar could be neat.

axes/models.py
arjenzijlstra: I agree, shall I remove it from the PR?

axes/tests/test_access_attempt.py
arjenzijlstra: I agree. I took it from another test and figured it was the style of this repos. I will rewrite it.

axes/tests/test_access_attempt.py
arjenzijlstra: I refactored the tests into a common function, I hope you like the way I did it. Let me know if I can do something in a better way.

axes/models.py
aleksihakli: Either is OK, just making a note :)

axes/tests/test_access_attempt.py
aleksihakli: Looks good to me.

aleksihakli:
Looks good, merging into upstream.

Issue #336, opened 2018-04-22
DjangoAxesModelBackend requires calls to authenticate to pass `request`


flexpeace:
I am getting this error DjangoAxesModelBackend requires calls to authenticate to pass `request` after following the instructions. 

Could you guys help me out?

Thanks

aleksihakli:
`DjangoAxesModelBackend` is not a backend in this project. Are you talking about `AxesModelBackend`?

What are your Django `AUTHENTICATION_BACKENDS`?

Which version of Axes are you using?

Have you read the [related documentation](https://django-axes.readthedocs.io/en/latest/configuration.html#authentication-backend-problems)?

PR #369, opened 2018-11-06
Update message of parameter required exception in AxesModelBackend.

arjenzijlstra:
Also mentioned in #336. I noticed that the error message refers to `DjangoAxesModelBackend`, while the class is called just `AxesModelBackend`.


markddavidoff:
Thanks!

PR #375, opened 2018-12-18
M3 support

ramidk:
Hello! I added:
1) FAILURE_LIMIT_MAX_BY_USER, COOLOFF_TIME_FORMATTER_CALLABLE settings
2) More configurable error messaging.
3) Support for M3 platform. (closed source software). 

All changes are backward compatible.


markddavidoff:
Please if possible split these up into 3 prs. Thanks

aleksihakli:
Hello @ramidk, and thank you for the pull request and the effort you have put into the axes project.

There are a few things that need to be improved, in my opinion, in order to make this work compatible with axes:

- Make the PRs smaller, this is a huge chunk and contains refactoring, new features, logic changes, translations in a new language, and proprietary integrations. These all should ideally be their own PRs.
- The checks should live in a separate module instead of `apps.py`. Something like `checks.py` that is called from `apps.py` would be good.
- Implement the M3 backend as a plugin based solution instead of hardcoded, so that you can separate the M3 backend into it's own repo and package. It's quite proprietary and the M3 specific code does not belong into the axes core.

The biggest issue here is that you are introducing hardcoded proprietary changes into a public, common project. If everybody did that, this project would become unmaintainable. 

The correct solution is to implement pluggability into axes and implement the proprietary stuff as plugins.

I'll leave it up to you to figure out a good plugin mechanism.

Issue #358, opened 2018-09-28
Should axes use the request.POST data or authenticate's **credentials?


markddavidoff:
In my use case, I process the POST to get the user we pass to `authenticate` from different endpoints in different ways. This is a hassle to reproduce in `AXES_USERNAME_CALLABLE`. I think instead, we should pull username and password from the args passed to authenticate instead of from the `request.POST`. Furthermore, request being present depends on it being correctly passed by 3rd party backends even though Django mentions it is an optional parm in the docs.

Input? @camilonova @aleksihakli @aclark4life @codekoala

markddavidoff:
It's also weird that we do `request.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)` in some places and `credentials[settings.AXES_USERNAME_FORM_FIELD]` in others.

markddavidoff:
I am willing to open a PR with changes to use credentials instead of request.POST for username fields but this would be a big-ish change so i want to know if people would be on board with it.

camilonova:
Using credentials sounds good to me.

mastacheata:
I'd like to suggest considering this as a bug.
Two reasons for this:
#1 Inconsistent behavior between storing the access attempt and looking it up.
#2 What if you send the username in a different way than as a POST form field?

Especially #2 makes this a huge PITA for me. We're submitting login credentials via JSON inside the POST body. Now Axes has no problem with storing the right stuff to the database, but if we use Username or Username+UA as the main identifier instead of the IP address, this will cause Axes to create new entries in the axes_accessattempts table instead of updating an existing record and increasing the failed login count. It tries to perform a lookup based on None as the username and that obviously fails.

aleksihakli:
@mastacheata good consideration, added the bug label.

aleksihakli:
Fixed by #377 and #380.

PR #377, opened 2018-12-19
Bugfix #358 credentials in favor of request

### LINKED ISSUE #376: Lock by username in DRF
Seems that lock user by username doesn't work in drf. I set `AXES_ONLY_USER_FAILURES=True` option in my settings file. I'm using standard drf token authentication, when I'm sending post request for obtain token I check users lock status using `is_already_locked` function, in this function axes generates cache key from username passed in post request, axes looks for this username in `request.POST`(https://github.com/jazzband/django-axes/blob/master/axes/attempts.py#L64), but in case of drf `request.POST` will be empty because drf store request payload in `request.data`. I tried to create custom function for retrieving username from request using `AXES_USERNAME_CALLABLE` option but function that generate cache key ignores my custom function and reads username directly from `request.POST`.

mastacheata:
As previously mentioned in #358 the `AXES_ONLY_USER_FAILURES` setting only works if the login is performed using a Form request.
If logging in using JSON, Axes can properly save the failed logins, but is unable to find an existing entry in the `axes_attempts` table or in cache due to relying on `request.POST` to find the username.

I changed this to make use of the credentials or username fields wherever possible, added new tests for thsi behavior and updated the documentation.

When available, `credentials` is passed along until it reaches the `get_client_username` in utils.py.
If `credentials` is not available to be forwarded, but the parsed username from login is available otherwise, a dummy `credentials` dictionary is created with the `settings.AXES_USERNAME_FORM_FIELD` key and the username inside it.
Only if none of these are availeble the old way to get the username directly from `request,POST` is attempted.
This mainly affects the decorators for the Login forms, but there's neither a good way to access the username otherwise nor will these usually be manipulated to not send a form request for the login.

Last but not least, I had to introduce a breaking change:
The `AXES_USERNAME_CALLABLE` function now receives the credentials dictionary as additional parameter. 

I think this should also fix #376 as I've just replaced the manual logic for looking up the username from the request with callng `get_client_username` instead

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -11,12 +11,12 @@ from axes.models import AccessAttempt
 from axes.utils import get_axes_cache, get_client_ip, get_client_username
 
 
-def _query_user_attempts(request):
+def _query_user_attempts(request, credentials):
     """Returns access attempt record if it exists.
     Otherwise return None.
     """
     ip = get_client_ip(request)
-    username = get_client_username(request)
+    username = get_client_username(request, credentials)
 
     if settings.AXES_ONLY_USER_FAILURES:
         attempts = AccessAttempt.objects.filter(username=username)
@@ -49,10 +49,11 @@ def _query_user_attempts(request):
     return attempts
 
 
-def get_cache_key(request_or_obj):
+def get_cache_key(request_or_obj, credentials=None):
     """
     Build cache key name from request or AccessAttempt object.
     :param  request_or_obj: Request or AccessAttempt object
+    :param  credentials: Dictionary with access credentials - Only supplied when request_or_obj is not an AccessAttempt
     :return cache-key: String, key to be used in cache system
     """
     if isinstance(request_or_obj, AccessAttempt):
@@ -61,8 +62,11 @@ def get_cache_key(request_or_obj):
         ua = request_or_obj.user_agent
     else:
         ip = get_client_ip(request_or_obj)
-        un = request_or_obj.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
         ua = request_or_obj.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
+        if credentials is None:
+            un = request_or_obj.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
+        else:
+            un = credentials.get(settings.AXES_USERNAME_FORM_FIELD, None)
 
     ip = ip.encode('utf-8') if ip else ''.encode('utf-8')
     un = un.encode('utf-8') if un else ''.encode('utf-8')
@@ -96,10 +100,10 @@ def get_cache_timeout():
     return cache_timeout
 
 
-def get_user_attempts(request):
+def get_user_attempts(request, credentials):
     force_reload = False
-    attempts = _query_user_attempts(request)
-    cache_hash_key = get_cache_key(request)
+    attempts = _query_user_attempts(request, credentials)
+    cache_hash_key = get_cache_key(request, credentials)
     cache_timeout = get_cache_timeout()
 
     cool_off = settings.AXES_COOLOFF_TIME
@@ -125,13 +129,13 @@ def get_user_attempts(request):
     # If objects were deleted, we need to update the queryset to reflect this,
     # so force a reload.
     if force_reload:
-        attempts = _query_user_attempts(request)
+        attempts = _query_user_attempts(request, credentials)
 
     return attempts
 
 
-def reset_user_attempts(request):
-    attempts = _query_user_attempts(request)
+def reset_user_attempts(request, credentials):
+    attempts = _query_user_attempts(request, credentials)
     count, _ = attempts.delete()
 
     return count
@@ -151,7 +155,7 @@ def ip_in_blacklist(ip):
     return ip in settings.AXES_IP_BLACKLIST
 
 
-def is_user_lockable(request):
+def is_user_lockable(request, credentials):
     """Check if the user has a profile with nolockout
     If so, then return the value to see if this user is special
     and doesn't get their account locked out
@@ -165,7 +169,7 @@ def is_user_lockable(request):
     try:
         field = getattr(get_user_model(), 'USERNAME_FIELD', 'username')
         kwargs = {
-            field: get_client_username(request)
+            field: get_client_username(request, credentials)
         }
         user = get_user_model().objects.get(**kwargs)
 
@@ -182,7 +186,7 @@ def is_user_lockable(request):
     return True
 
 
-def is_already_locked(request):
+def is_already_locked(request, credentials=None):
     ip = get_client_ip(request)
 
     if (
@@ -200,10 +204,10 @@ def is_already_locked(request):
     if ip_in_blacklist(ip):
         return True
 
-    if not is_user_lockable(request):
+    if not is_user_lockable(request, credentials):
         return False
 
-    cache_hash_key = get_cache_key(request)
+    cache_hash_key = get_cache_key(request, credentials)
     failures_cached = get_axes_cache().get(cache_hash_key)
     if failures_cached is not None:
         return (
@@ -211,7 +215,7 @@ def is_already_locked(request):
             settings.AXES_LOCK_OUT_AT_FAILURE
         )
 
-    for attempt in get_user_attempts(request):
+    for attempt in get_user_attempts(request, credentials):
         if (
             attempt.failures_since_start >= settings.AXES_FAILURE_LIMIT and
             settings.AXES_LOCK_OUT_AT_FAILURE
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -63,10 +63,7 @@ def get_cache_key(request_or_obj, credentials=None):
     else:
         ip = get_client_ip(request_or_obj)
         ua = request_or_obj.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
-        if credentials is None:
-            un = request_or_obj.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
-        else:
-            un = credentials.get(settings.AXES_USERNAME_FORM_FIELD, None)
+        un = get_client_username(request_or_obj, credentials)
 
     ip = ip.encode('utf-8') if ip else ''.encode('utf-8')
     un = un.encode('utf-8') if un else ''.encode('utf-8')
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -4,6 +4,7 @@ from django.contrib.auth.backends import ModelBackend
 from django.core.exceptions import PermissionDenied
 
 from axes.attempts import is_already_locked
+from axes.conf import settings
 from axes.utils import get_lockout_message
 
 
@@ -30,10 +31,13 @@ class AxesModelBackend(ModelBackend):
         :return: Nothing, but will update return_context with lockout message if user is locked out.
         """
 
+        # Create credentials dictionary from username field
+        credentials = {settings.AXES_USERNAME_FORM_FIELD: username}
+
         if request is None:
             raise AxesModelBackend.RequestParameterRequired()
 
-        if is_already_locked(request):
+        if is_already_locked(request, credentials):
             # locked out, don't try to authenticate, just update return_context and return
             # Its a bit weird to pass a context and expect a response value but its nice to get a "why" back.
             error_msg = get_lockout_message()
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -38,7 +38,7 @@ def log_user_login_failed(sender, credentials, request, **kwargs):  # pylint: di
         return
 
     ip_address = get_client_ip(request)
-    username = get_client_username(request)
+    username = get_client_username(request, credentials)
     user_agent = request.META.get('HTTP_USER_AGENT', '<unknown>')[:255]
     path_info = request.META.get('PATH_INFO', '<unknown>')[:255]
     http_accept = request.META.get('HTTP_ACCEPT', '<unknown>')[:1025]
@@ -47,8 +47,8 @@ def log_user_login_failed(sender, credentials, request, **kwargs):  # pylint: di
         return
 
     failures = 0
-    attempts = get_user_attempts(request)
-    cache_hash_key = get_cache_key(request)
+    attempts = get_user_attempts(request, credentials)
+    cache_hash_key = get_cache_key(request, credentials)
     cache_timeout = get_cache_timeout()
 
     failures_cached = get_axes_cache().get(cache_hash_key)
@@ -110,7 +110,7 @@ def log_user_login_failed(sender, credentials, request, **kwargs):  # pylint: di
     if (
         failures >= settings.AXES_FAILURE_LIMIT and
         settings.AXES_LOCK_OUT_AT_FAILURE and
-        is_user_lockable(request)
+        is_user_lockable(request, credentials)
     ):
         log.warning(
             'AXES: locked out %s after repeated login attempts.',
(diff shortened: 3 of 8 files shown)


docs/usage.rst
```diff
@@ -96,7 +96,7 @@ them as per the example.
                 request = request,
                 username = form.cleaned_data.get('username'),
                 password = form.cleaned_data.get('password'),
-            ) 
+            )
```
aleksihakli: Unrelated to this PR.

docs/usage.rst
```diff
@@ -129,22 +129,22 @@ Integration with django-allauth
...
 
-You also need to decorate ``dispatch()`` and ``form_invalid()`` methods 
-of the ``allauth`` login view. By default ``axes`` is patching only the 
+You also need to decorate ``dispatch()`` and ``form_invalid()`` methods
+of the ``allauth`` login view. By default ``axes`` is patching only the
 ``LoginView`` from ``django.contrib.auth`` app and with ``allauth`` you have to
 do the patching of views yourself.
 
 *settings.py:* ::
-
```
aleksihakli: Unrelated to this PR.

docs/usage.rst
```diff
@@ -129,22 +129,22 @@ Integration with django-allauth
...
+``user_login_failed`` signal. This is not the case with ``allauth``.
 ``allauth`` always uses ``login`` key in post POST data but it becomes ``username``
 key in ``credentials`` dict in signal handler.
 
 To overcome this you need to use custom login form that duplicates the value
-of ``username`` key under a ``login`` key in that dict 
+of ``username`` key under a ``login`` key in that dict
 (and set ``AXES_USERNAME_FORM_FIELD = 'login'``).
 
-You also need to decorate ``dispatch()`` and ``form_invalid()`` methods
```
aleksihakli: Unrelated to this PR.

docs/usage.rst
```diff
@@ -129,22 +129,22 @@ Integration with django-allauth
...
 ``axes`` relies on having login information stored under ``AXES_USERNAME_FORM_FIELD`` key
-both in ``request.POST`` and in ``credentials`` dict passed to 
-``user_login_failed`` signal. This is not the case with ``allauth``. 
+both in ``request.POST`` and in ``credentials`` dict passed to
+``user_login_failed`` signal. This is not the case with ``allauth``.
 ``allauth`` always uses ``login`` key in post POST data but it becomes ``username``
 key in ``credentials`` dict in signal handler.
 
 To overcome this you need to use custom login form that duplicates the value
-of ``username`` key under a ``login`` key in that dict
```
aleksihakli: Unrelated to this PR.

docs/usage.rst
```diff
@@ -129,22 +129,22 @@ Integration with django-allauth
 -------------------------------
 
 ``axes`` relies on having login information stored under ``AXES_USERNAME_FORM_FIELD`` key
-both in ``request.POST`` and in ``credentials`` dict passed to
```
aleksihakli: Unrelated to this PR.

mastacheata:
No hurry.

Sure, if that's the way you prefer it, I'll separate out the whitespace changes.
I'm used to fixing minor style issues in one go with other changes where I work. But I can understand the opposite opinion as well. :D

markddavidoff:
we'll all proably have more time to look at this next year so thanks for your patience. 

My intial note was that the addition of passed param to so many functions indicates a design issue. I think this calls for an `AxesLoginAttemptData` (or whatever name) class that wraps the data we need into a class and can be easily passed to all the places we need it. This way future changes to the data we need don't require function definition changes everywhere. Open to arguments against, but i think that would help neaten things up a bit.

markddavidoff:
I'll also have more time to look at this next year so thanks for your patience. 

My intial note was that the addition of passed param to so many functions indicates a design issue. I think this calls for an `AxesLoginAttemptData` (or whatever name) class that wraps the data we need into a class and can be easily passed to all the places we need it. This way future changes to the data we need don't require function definition changes everywhere. Open to arguments against, but i think that would help neaten things up a bit.

axes/utils.py
```diff
@@ -69,10 +69,12 @@ def get_client_ip(request):
     return getattr(request, client_ip_attribute)
 
 
-def get_client_username(request):
+def get_client_username(request, credentials=None):
     if settings.AXES_USERNAME_CALLABLE:
-        return settings.AXES_USERNAME_CALLABLE(request)
-    return request.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
+        return settings.AXES_USERNAME_CALLABLE(request, credentials)
+    if credentials is None:
```
markddavidoff: just a semantics note, but reversing this logic to be `is not None` would make it slightly more clear that the request method is the "fallback"

docs/configuration.rst
```diff
@@ -124,10 +124,10 @@ These should be defined in your ``settings.py`` file.
   Default: ``True``
 * ``AXES_USERNAME_FORM_FIELD``: the name of the form field that contains your
   users usernames. Default: ``username``
-* ``AXES_USERNAME_CALLABLE``: A callable function that takes a request object
-  and returns the username. If empty, axes just fetches it from the POST body
+* ``AXES_USERNAME_CALLABLE``: A callable function that takes a request object and a credentials dictionary
```
markddavidoff: "credentials dictionary" is not very clearly defined. we should be more explicit.

axes/utils.py
```diff
@@ -69,10 +69,12 @@ def get_client_ip(request):
     return getattr(request, client_ip_attribute)
 
 
-def get_client_username(request):
+def get_client_username(request, credentials=None):
     if settings.AXES_USERNAME_CALLABLE:
-        return settings.AXES_USERNAME_CALLABLE(request)
-    return request.POST.get(settings.AXES_USERNAME_FORM_FIELD, None)
+        return settings.AXES_USERNAME_CALLABLE(request, credentials)
```
markddavidoff: changelog need to indicate that adding this extra param will be a breaking change and require people to update their existing AXES_USERNAME_CALLABLE  funcs

axes/backends.py
```diff
@@ -30,10 +31,13 @@ def authenticate(self, request, username=None, password=None, **kwargs):
         :return: Nothing, but will update return_context with lockout message if user is locked out.
         """
 
+        # Create credentials dictionary from username field
+        credentials = {settings.AXES_USERNAME_FORM_FIELD: username}
```
markddavidoff: lets pass all the authenticate params here including extra params. that would be useful and consistently formated for people using their own callables. WE don't need the "settings.AXES_USERNAME_FORM_FIELD" key as we already know the username will be in the 'username' param. (make sure to update the usage in def get_client_username to match too)

axes/signals.py
```diff
@@ -148,7 +148,9 @@ def log_user_logged_in(sender, request, user, **kwargs):  # pylint: disable=unus
         )
 
     if settings.AXES_RESET_ON_SUCCESS:
-        count = reset_user_attempts(request)
+        # Create credentials dictionary from the username field
+        credentials = {settings.AXES_USERNAME_FORM_FIELD: username}
```
markddavidoff: why overwrite credentials?

axes/attempts.py
```diff
@@ -49,10 +49,11 @@ def _query_user_attempts(request):
     return attempts
 
 
-def get_cache_key(request_or_obj):
+def get_cache_key(request_or_obj, credentials=None):
     """
     Build cache key name from request or AccessAttempt object.
     :param  request_or_obj: Request or AccessAttempt object
+    :param  credentials: Dictionary with access credentials - Only supplied when request_or_obj is not an AccessAttempt
```
markddavidoff: same note as from doc.

aleksihakli:
I agree with @markddavidoff and think that the passing of multiple arguments throughout the call stack is a sign of bad call signature design. We should make an argument object type that can be instantiated once and passed through the whole call stack after that, as that would make everything considerably neater and more testable if implemented correctly. What if we need to add one more argument to indicate some kind of order of preference or other metadata? Shall we alter 50 function signatures, or just augment a class?

If we implement a custom credentials wrapper class I think we are free to do it as we like, but the Django docs should be referenced on what the call signatures in the core are to find out if we can be compliant with the core concepts and ideas of Django in structuring axes.

mastacheata:
I'll look into most of the comments tomorrow afternoon.

@aleksihakli I'm not sure how to handle that last comment of yours.
At first you say I should better build a single wrapper object for all the arguments that need to be passed along, but most of the time I'm passing both the request AND the credentials along.
In the next paragraph you then talk about wrapping the credentials in an own object/class instead of a simple dictionary like the native django authentication does.

I think the request object is of high enough relevance to be passed along "as is" and would only wrap the credentials or fake credentials dictionary in a new object for now as the request is used as the fallback/defacto default.

axes/signals.py
mastacheata: I'm not overwriting them here.
Did you maybe add that comment to the wrong line?
log_user_logged_in doesn't receive the credentials object from the authentication backend, but is only passed a user object. (which is not identical to the credentials dictionary)

axes/backends.py
mastacheata: Fine with the first remark. but the second seems wrong.
We know the username will be coming from the username function argument, but we don't know if the custom parsing function or the custom fieldname for all the other ways an authentication could be logged in AXES is expected to be username.
The get_client_username function is used in many places and completely oblivious to where it got called from.

We know Django Admin always uses username as the field and the default authentication backends do the same.
But what if someone uses the default Django Admin and a custom JSON supplied authentication in the same project?
The get_cleint_username would have to check both the credentials['username'] from Django Admin and the credentials[settings.AXES_USERNAME_FORM_FIELD] in such a case.

docs/configuration.rst
mastacheata: I changed the documentation (did incorporate the text from django docs, but also changed the wording to better fit inside a "wrapping sentence")

markddavidoff:
The plan is to deprecate the form and make  backends the way to process
this. Might as well start here.

axes/utils.py
mastacheata: Definitely.  
Should I change that or will one of you adjust the changelog when you release the next version with this fix included?

axes/signals.py
markddavidoff: you're right i was looking at the wrong line

aleksihakli:
@markddavidoff after consideration I'm wondering if we could merge this change set and start work on the new authentication wrapper after that? We could offer a clean API and refactor the internal implementation without affecting the external APIs that much, considering that the client side is using only the signal handlers, the authentication backend and the decorators for view and form dispatching, which will not be changing.

If we did that, we could merge this PR and offer better support for users for the time being in version 4 while working to implement the new internal APIs and release a version 5 with breaking changes after that?

If we support the custom `AXES_USERNAME_CALLABLE` backwards compatibility with a dynamic `inspect.signature` dispatching we can implement this PR without breaking anything for the time being.

The case where something breaks for current users is when somebody is using the internal APIs and the signatures in that implementation change, which isn't nice, but that can't be avoided anyway, if we want to implement the new facade wrappers for the authentication data.

markddavidoff:
@aleksihakli lots of good points. This pr does  have the one breaking change for the callable function though, so my motivation was just to group those 2 together, but I'm fine with going with smaller iterations as well. the middle ground would be to have a  feature branch or hold off on cutting a release until its all in. @mastacheata thanks again for the great work.

axes/backends.py
aleksihakli: There should probably be a `get_credentials` method that can be overridden here.

`username` argument is not always supplied here, actually. The argument might be `token` or something similar as well, it's up to the user. This signature just comes from the `ModelBackend` base class, but the Django authentication framework just specifies that the signature is `authenticate(request=None, **credentials)`.

axes/tests/test_utils.py
```diff
@@ -171,3 +186,22 @@ def test_custom_get_client_username(self):
         actual = get_client_username(request)
 
         self.assertEqual(expected, actual)
+
+    def sample_customize_username_from_credentials(request, credentials):
+        return 'prefixed-' + credentials.get('username')
+
+    @override_settings(AXES_USERNAME_FORM_FIELD='username')
```
aleksihakli: You can override multiple settings inside one `@override_settings` call.

axes/utils.py
aleksihakli: I agree, it would make this more readable as a whole.

docs/configuration.rst
```diff
@@ -124,10 +124,14 @@ These should be defined in your ``settings.py`` file.
   Default: ``True``
 * ``AXES_USERNAME_FORM_FIELD``: the name of the form field that contains your
   users usernames. Default: ``username``
-* ``AXES_USERNAME_CALLABLE``: A callable function that takes a request object
-  and returns the username. If empty, axes just fetches it from the POST body
+* ``AXES_USERNAME_CALLABLE``: A callable function that takes two arguments:
```
aleksihakli: I think this should be 

```
An optional custom callable that takes either one or two arguments. The supplied arguments are the request and optionally the credentials dictionary that was passed to `authenticate()` by Django. The function must return an username.
```

docs/usage.rst
```diff
@@ -187,15 +187,16 @@ removing client-set prefixes. In these cases, ``axes`` needs to know how to make
 these changes so that it can correctly identify the user without any form
 cleaning or validation. This is where the ``AXES_USERNAME_CALLABLE`` setting
 comes in. You can define how to make these modifications in a callable that
-takes a request object, and provide that callable to ``axes`` via this setting.
+takes a request object and a credentials dictionary,
```
aleksihakli: `and an optional credentials dictionary.`

aleksihakli:
After more consideration I think we can merge this to `master` and add the backwards compatibility fixes and inspections to `master` for the `AXES_USERNAME_CALLABLE` and other callables, so we don't break the external facing APIs. 

We can then make a new minor version release (https://semver.org/) `4.5.0` and add a release note about internal APIs changing slightly.

The larger refactoring of the data structures can then be done after this is upstreamed - it is probably even easier to refactor after the APIs are changed to the `(request, credentials)` parameter signature since then we have a basis for implementing the multiple parameter ingesting facade.

Do you have any notes on that @markddavidoff? Do you consider using `master` here OK with the backwards compatibility fixes, and then working on the new backwards incompatible API in an another feature branch, for example `facade`?

I can make the backwards compatibility fixes after the merge.

mastacheata:
I'll look into the minor changes (i.e. @override_settings, documentation) now and leave the real logic untouched for now until you've come to a conclusion on how to proceed.

aleksihakli:
Yeah, let's do that @mastacheata. This is good work, and I think it is well scoped. It just needs to be backwards compatible if we don't want to cut out multiple major releases, which is unconvenient for the users of this package. Let's wait for @markddavidoff comments on the subject so that we have a consensus on how to proceed with the feature and release planning.

mastacheata:
If you'd like to know my opinion, though: 
I'd favor the staged approach as well.
Basically this PR is a bugfix and only has the breaking change as an (unintentional) side effect.
I only wrote my first line of Python less than 3 months ago and couldn't think of an elegant way to have optional arguments supplied at the time of the function call rather than relying on having them in the function definition.

Also: Upgrading to a new minor release of a library is easy and done often even in larger projects, but upgrading to new major releases requires way more effort on that end.

aleksihakli:
@mastacheata this is now released to PyPI as version 4.5.0. Have a nice Xmas!

mastacheata:
WOW!
Had enough of all the christmas festivities? ;)
Thanks a lot for bringing this up to speed and implementing all those remaining changes.

Will definitely keep a look out at that new issue and the refactoring/redesign of this as well.

PR #378, opened 2018-12-19
Run tests in parallel in Travis

aleksihakli:
Attempting to produce a parallel build for Python - Django version combinations as described here:

- https://docs.travis-ci.com/user/speeding-up-the-build/
- https://tox-travis.readthedocs.io/en/stable/envlist.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,33 @@
-sudo: false
+dist: xenial
 language: python
 cache: pip
+python:
+  - 2.7
+  - 3.4
+  - 3.5
+  - 3.6
+  - 3.7
+env:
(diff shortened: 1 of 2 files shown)


aleksihakli:
Slightly improved display and speed of Travis builds for different Django and Python versions in the matrix: 

image

The build also removes the old deprecated `sudo: false` keyword, cleans up the `django-NNN` versions to `djangoNNN` syntax as recommended by the Travis and tox documentations and explicitly moves the builds to the new `xenial` build platform, which is the default platform starting from December 7th onwards:

https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

PR #379, opened 2018-12-19
Strip trailing spaces in usage.rst

mastacheata:
Trailing spaces in usage.rst removed
Single trailing spaces serve no purpose at all and double trailing spaces on a blank/separator line has no effect either.


markddavidoff:
sure.

PR #391, opened 2019-01-12
Turkish translation added

obayhan:
Turkish translation added (mo and po files)


aleksihakli:
Thank you for the contribution @obayhan. I reviewed the changes with a few translator programs and they seem to be good, so I'm merging and releasing this into upstream.

PR #402, opened 2019-02-16
Refactor layout to allow better pluggability and specialized handlers in the authentication stack

aleksihakli:
There are a lot of changes in this PR but most of them focus in the `axes.utils` and `axes.attempts` modules which I refactored in order to clean up their implementations a bit.

In the same work I added some better settings with the `settings.AXES_CLIENT_IP_ATTRIBUTE` being settable and configurable for users, `settings.AXES_USERNAME_CALLABLE` supporting string attributes as well as callables and `settings.AXES_HANDLER` supporting reloading in tests and reconfigurable runtimes with the [Django settings changed signals](https://docs.djangoproject.com/en/dev/ref/signals/#setting-changed).

I'll write a thorough migration instruction set and documentation once we have the API sketched out properly, but I think doing that is smarter after we have the new cached handler in the project.


aleksihakli:
As nobody has opposed this restructuring and API design during this week while this changeset has been open for review, I am merging this changeset into upstream. 

We can take a more comprehensive look into the overall API restructuring before releasing version 5, but this pluggable signal handler stack is an overall prerequisite and technical requirement for implementing the cached handlers that are central part of the Django Defender design in Django Axes.

If somebody opposes these changes later on, please reopen discussion here or contact me. If something looks like it breaks backwards compatibility or deprecates an API that is needed, we can work further on compatibility and internal APIs.

Thanks to everyone who has offered public and private comments on the design and implementation of this new layer of functionality. Much of the work done here is not in this changeset and commits and is not visible in the git log, but is greatly valued.

axes/handlers/base.py
```diff
@@ -0,0 +1,54 @@
+from typing import Any, Dict, Optional
+
+from django.http import HttpRequest
+
+
+class AxesBaseHandler:  # pylint: disable=unused-argument
```
jorlugaqui: @aleksihakli I think this class could be [abstract](https://docs.python.org/3.7/library/abc.html ), seems to me that every specialized handler should implement all the methods defined in here. Also it would tell to developers clearly that this base class pretends to define the general API.

axes/handlers/database.py
```diff
@@ -0,0 +1,283 @@
...
+    def user_login_failed(self, sender, credentials, request, **kwargs):  # pylint: disable=too-many-locals
+        """
+        When user login fails, save AccessAttempt record in database and lock user out if necessary.
+
+        :raises AxesSignalPermissionDenied: if user should is locked out
+        """
+
+        if request is None:
+            log.warning('AXES: AxesDatabaseHandler.user_login_failed does not function without a request.')
+            return
```
jorlugaqui: @aleksihakli out of curiosity. I'm trying to remember but I can't;  why are not we returning an exception for this cases?

axes/handlers/database.py
```diff
@@ -0,0 +1,283 @@
...
+        get_data = get_query_str(request.GET)
+        post_data = get_query_str(request.POST)
+        attempt_time = now()
+
+        if is_client_ip_address_whitelisted(request):
+            log.info('AXES: Login failed from whitelisted IP %s.', ip_address)
+            return
+
+        attempts = get_user_attempts(request, credentials)
+        cache_key = get_cache_key(request, credentials)
```
jorlugaqui: @aleksihakli I think these two lines could be somehow moved to a helper, I've seen them in many places. Maybe in `utils.py`, something like?

```
num_failures_cached = get_cached_failures(request, credentials)
...
get_cached_failures(request, credentials):
    cache_key = get_cache_key(request, credentials)
    return get_axes_cache().get(cache_key)
```

jorlugaqui: @aleksihakli also, if the attempts were found in the cache, I think we don't need to check for those in the database.

jorlugaqui: @aleksihakli ok, never mind, I noticed we would need the attempts anyway.

axes/models.py
```diff
@@ -56,12 +56,8 @@ class AccessAttempt(AccessBase):
         _('Failed Logins'),
     )
 
-    @property
-    def failures(self):
-        return self.failures_since_start
-
     def __str__(self):
-        return 'Attempted Access: %s' % self.attempt_time
+        return 'Attempted Access: {}'.format(self.attempt_time)
```
jorlugaqui: @aleksihakli fstrings maybe?

jorlugaqui:
@aleksihakli thank you for your work and sorry for checking it that late. I have left some comments that by any means pretend to reopen this ticket. If they apply in your opinion, I could create a PR.

jorlugaqui:
@aleksihakli also keep in mind that some comments might not apply because of your changes on #405.

axes/handlers/base.py
aleksihakli: That's a good idea, I'll see into the abstractness.

axes/handlers/database.py
aleksihakli: This is the previous behaviour and I opted to go with backwards compatible fashion. Raising would probably be the correct way to behave here, though, I agree on that.

axes/handlers/database.py
aleksihakli: The database caching was removed altogether, so this is a bit dated place, but fetching the failures would need some refactoring :)

axes/models.py
aleksihakli: f-strings are only available in Python 3.6+ and we're still supporting Python 3.5 and PyPy 3.5 so I opted for the old-fashion `.format`. It should be easy enough to migrate to f-strings when the time is ripe and Python 3.5 support is removed.

aleksihakli:
Thanks for the comments and the review! If you have suitable improvements in mind, I think PR would be a good idea. Having a fresh set of eyes go over the work on master branch would be very nice.

axes/handlers/database.py
jorlugaqui: cc @camilonova any thought on that?

axes/models.py
jorlugaqui: @aleksihakli oh, you are right, don't know why I thought we will support >= 3.6.

Issue #401, opened 2019-02-14
Add cached backend handler


jorlugaqui:
As per the conversation happening in https://github.com/kencochrane/django-defender/issues/121#issuecomment-461617352, this issue aims for getting a pluggable cached backend for django-axes. 

If you don't mind, I could give it a try.

aleksihakli:
@jorlugaqui sounds awesome, if you have the time to spare I would greatly appreciate the help! Just go ahead and open a PR and let's see how things turn out.

I am also playing around the current middleware - backend - handler structure and thinking on how to clean it up so that the locking checks would be simpler. Big part of this work is in the refactoring of the current implementation.

I think that working on the purely cached backend solution on a parallel track would only be beneficial for both tracks.

It could also help us possibly improve the current API structure before freezing and releasing it in Axes version 5. We can't really change things around after baking that release, but have plenty of flexibility at this time, so go ahead and propose the changes you would like to see if you have any ideas.

aleksihakli:
@jorlugaqui @camilonova I think that the handler layer could benefit from moving the Axes internal method `axes.attempts.is_already_locked` into the handler class.

The `is_already_locked` function being in the handler layer would allow us to just change the handler implementation for checking the lockout status, and using the default Axes authentication backend and middleware for the lockouts.

The handler could then execute the lockout checks on the correct layer (database or cache or e.g. static ruleset).

The `is_already_locked` method is currently called from the decorators, the authentication backend and the signal handlers in the current implementation, once during the authentication backend which checks if the user is allowed to log in and another time in the signal handler which checks if the user should be locked out:

- https://github.com/jazzband/django-axes/blob/ad2ed23ced5590d8a936f3194a895dd69d3e5707/axes/decorators.py#L9
- https://github.com/jazzband/django-axes/blob/ad2ed23ced5590d8a936f3194a895dd69d3e5707/axes/decorators.py#L20
- https://github.com/jazzband/django-axes/blob/ad2ed23ced5590d8a936f3194a895dd69d3e5707/axes/backends.py#L39
- https://github.com/jazzband/django-axes/blob/ad2ed23ced5590d8a936f3194a895dd69d3e5707/axes/handlers.py#L117

The implementation would be fairly easy to move, I think:

- https://github.com/jazzband/django-axes/blob/ad2ed23ced5590d8a936f3194a895dd69d3e5707/axes/attempts.py#L188

Just hoist that function implementation into the handler class.

[...]

PR #405, opened 2019-02-22
Add cache handler

aleksihakli:
Cache handler as discussed in #401.

This changeset moves the caching handler logic from the database handler to the cache handler, allowing for cleaner and easier code structuring and separation of concerns. The two main places of interest are the `axes.handlers.database.AxesDatabaseHandler` and `axes.handlers.cache.AxesCacheHandler`.

@jorlugaqui and @joshua-s what do you think of this approach?


axes/handlers/cache.py
```diff
@@ -0,0 +1,116 @@
...
+        if request is None:
+            log.error('AXES: AxesCacheHandler.user_login_failed does not function without a request.')
+            return
+
+        username = get_client_username(request, credentials)
+        ip_address = get_client_ip_address(request)
+        user_agent = get_client_user_agent(request)
+        path_info = get_client_path_info(request)
+        client_str = get_client_str(username, ip_address, user_agent, path_info)
+
```
joshua-s: I wonder if these could be placed into a single helper function

axes/handlers/cache.py
```diff
@@ -0,0 +1,116 @@
...
+                client_str,
+            )
+
+        cache = get_axes_cache()
+        cache_key = get_client_cache_key(request, credentials)
+        cache_timeout = get_cool_off().total_seconds()
+
+        cache.set(cache_key, failures_since_start, cache_timeout)
+
+        if failures_since_start >= settings.AXES_FAILURE_LIMIT:
```
joshua-s: Since the `failures_since_start` logic seems to apply to both handlers, you could move it into a class method so it wouldn't need to be written out unless a handler wanted to override it.

axes/handlers/database.py
```diff
@@ -42,69 +35,30 @@ class AxesDatabaseHandler(AxesBaseHandler):  # pylint: disable=too-many-locals
...
-        3. Is the request user _whitelisted_? If it is, return ``True``.
-        5. Is failed authentication attempt always allowed to proceed? If it is, return ``True``.
-        6. Is failed authentication attempt count over the attempt limit? If it is, return ``False``.
-
-        Refer to the function source code for the exact implementation.
-        """
-
-        if is_client_ip_address_blacklisted(request):
+    def is_locked(self, request, credentials=None, attempt_time=None):
+        if is_user_attempt_whitelisted(request, credentials):
```
joshua-s: Is this something that might be useful if it were moved to the parent function?

axes/handlers/cache.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class AxesCacheHandler(AxesBaseHandler):  # pylint: disable=too-many-locals
+    """
+    Signal handler implementation that records user login attempts to cache and locks users out if necessary.
+    """
+
+    def get_failures(self, request, credentials=None, attempt_time=None) -> int:
+        cache = get_axes_cache()
+        cache_key = get_client_cache_key(request, credentials)
```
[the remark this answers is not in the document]
aleksihakli: Yeah, that's something I would like to do on a refactoring round. Many of the utility functions and loaders could be lazier.

axes/handlers/cache.py
aleksihakli: They could and most probably should. One good place to load and set these would be the middleware, but I was considering if that is the best place since some people could opt to use their own middleware or skip the Axes one. 

Fetching these where they are used imposes the least state, and we would ideally like this handler chain to be as stateless as possible, so a single helper function is probably the best place.

axes/handlers/cache.py
aleksihakli: Yeah, these checks should ideally be refactored so that they are in common places in the base implementation.

axes/handlers/database.py
aleksihakli: I was thinking about the `is_user_attempt_whitelisted` for a while: the reason it is here is that it touches the database and checks the user model. 

If the cache handler should be purely cache based and not impose any load on the database on higher traffic sites, then having this in the cache backend as well (via the base class) produces redundant queries in all inheriting backends by default.

I was altogether considering removing this check as a whole since it seems to be a quite specialized check and should, at the very least, be more configurable than this. 

On the other hand, the handler based approach does offer a fairly customizable base for implementations, so maybe we are on good grounds here.

aleksihakli:
Hi, and thanks for the comments and feedback! 

I was considering leaving the PR open for a while longer, but came to the conclusion that it's maybe better to just go ahead and upstream it so people can further work on the handlers.

The implementations are fairly well tested, though the unit and integration tests need some improvement for readability and reusability. Making new handlers very easy to unit test with Axes would be a thing that would be worth pursuing.

I'm still planning to improve the handler design and implementations a bit and work on the documentation, and possibly add a cached database handler to further refine the design, and would appreciate any help on honing the design there :)

axes/handlers/cache.py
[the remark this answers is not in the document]
aleksihakli: Implemented as a class instance attribute in https://github.com/jazzband/django-axes/commit/7564603d9bc2ca8410dda738994706634e0f3d1d

aleksihakli:
Upstream documentation is now updated with information about handlers and the new architecture.

axes/attempts.py
```diff
@@ -62,35 +43,35 @@ def filter_user_attempts(request: HttpRequest, credentials: dict = None) -> Quer
...
+    Get valid user attempts that match the given request and credentials.
     """
 
     attempts = filter_user_attempts(request, credentials)
 
-    # If settings.AXES_COOLOFF_TIME is not configured return the attempts
-    cool_off = get_cool_off()
-    if cool_off is None:
+    if settings.AXES_COOLOFF_TIME is None:
+        log.debug('AXES: Getting all access attempts from database because no AXES_COOLOFF_TIME is configured')
```
jorlugaqui: @aleksihakli are those all `log.debug` be removed or updated to log.info?

axes/attempts.py
```diff
@@ -99,6 +80,55 @@ def reset_user_attempts(request: HttpRequest, credentials: dict = None) -> int:
...
+
+    This utility method is meant to be used from the CLI or via Python API.
+    """
+
+    attempts = AccessAttempt.objects.all()
+
+    if ip:
+        attempts = attempts.filter(ip_address=ip)
+    if username:
+        attempts = attempts.filter(username=username)
```
jorlugaqui: @aleksihakli correct me if I'm wrong, but in case both `ip` and `username` are present, this will perform an `and` and not and `or` operation. So maybe in the docs would be worthy to mention this.

axes/attempts.py
```diff
@@ -99,6 +80,55 @@ def reset_user_attempts(request: HttpRequest, credentials: dict = None) -> int:
...
+
+
+def is_user_attempt_whitelisted(request: HttpRequest, credentials: dict = None) -> bool:
+    """
+    Check if the given request or credentials refer to a whitelisted username.
+
+    A whitelisted user has the magic ``nolockout`` property set.
+
+    If the property is unknown or False or the user can not be found,
+    this implementation fails gracefully and returns True.
```
jorlugaqui: @aleksihakli I could not see where is returning True.

axes/handlers/cache.py
```diff
@@ -0,0 +1,116 @@
...
+                failures_since_start,
+                settings.AXES_FAILURE_LIMIT,
+            )
+        else:
+            log.warning(
+                'AXES: New login failure by %s. Creating new record in the cache.',
+                client_str,
+            )
+
+        cache = get_axes_cache()
```
jorlugaqui: @aleksihakli aren't those lines the same as `self.get_failures`?

axes/handlers/cache.py
```diff
@@ -0,0 +1,116 @@
...
+        credentials = get_credentials(username)
+        ip_address = get_client_ip_address(request)
+        user_agent = get_client_user_agent(request)
+        path_info = get_client_path_info(request)
+        client_str = get_client_str(username, ip_address, user_agent, path_info)
+
+        log.info('AXES: Successful login by %s.', client_str)
+
+        if settings.AXES_RESET_ON_SUCCESS:
+            cache = get_axes_cache()
```
jorlugaqui: @aleksihakli aren't these lines the same as `self.get_failures`?

axes/tests/settings.py
```diff
@@ -7,7 +10,11 @@
 
 CACHES = {
     'default': {
-        'BACKEND': 'django.core.cache.backends.dummy.DummyCache'
+        'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
```
jorlugaqui: @aleksihakli I'm wondering why it would be better to have a file and not a dummy cache for the tests?

jorlugaqui:
@aleksihakli same as in #402, thanks for your work. I left a few comments that can be ignored. I did not have the chance of running the app, but will do it during the week so I can get more insights / questions/ concerns. Will let you know.

axes/attempts.py
aleksihakli: I think that these debug level outputs are useful for people who wish to gain more insight into their application, but are not that useful when running the application in normal production mode, hence the info level.

axes/attempts.py
aleksihakli: Good note, I'll upgrade the docs accordingly.

axes/handlers/cache.py
aleksihakli: Yeah, these methods still have some overlap and need some refactoring and cleanup. I'd just like to keep the delta between the readability of the database and cache backend small so they are easy to compare.

axes/tests/settings.py
aleksihakli: The dummy cache doesn't work in the cache tests since it doesn't record any of the failed attempts whereas the FileBasedCache was easy enough to configure for Travis compatibility. It is also suited for a single server deployments on Axes, so I thought it would be a good default in the test bench.

jorlugaqui: @aleksihakli right, right :)

PR #416, opened 2019-03-03
Set Axes request attributes in middleware

aleksihakli:
@jorlugaqui this PR implements the request attribute calculation in middleware as discussed previously.

### DIFF
--- a/axes/middleware.py
+++ b/axes/middleware.py
@@ -1,5 +1,15 @@
+from django.http import HttpRequest
+from django.utils.timezone import now
+
 from axes.exceptions import AxesSignalPermissionDenied
-from axes.helpers import get_lockout_response
+from axes.helpers import (
+    get_client_ip_address,
+    get_client_user_agent,
+    get_client_path_info,
+    get_client_http_accept,
+    get_lockout_response,
+)
+from axes.request import AxesHttpRequest
 
 
 class AxesMiddleware:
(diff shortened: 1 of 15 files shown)


axes/middleware.py
```diff
@@ -21,10 +31,16 @@ class AxesMiddleware:
     def __init__(self, get_response):
         self.get_response = get_response
 
-    def __call__(self, request):
+    def __call__(self, request: HttpRequest):
```
jorlugaqui: @aleksihakli very clever, I would have never thought about adding all those values to the request. My only recommendation would be to copy the request into an `AxesHttpRequest` object, and then attach the new attributes. Feels more natural and more semantic to me to send to an `AxesHttpRequest` to the `self.get_response` method.

aleksihakli: Yeah, I was considering if that would be the smarter way to do things. 

The `request` objects in the Django middleware stack are often mutated and attributes are added when processing the request, and I thought that doing a copy of the data adds overheads and might introduce strange bugs in some edge cases where the object state or reference to the request might actually be important, so I opted to go with adding the attributes and adding a signature to the places where the `request` is used.

This is less semantically valid, but might be more in line with existing practices of using the `request` as a store for all the related data.

I thought I'd go through the code bases of some competing implementations and established Django plugins with existing clean middleware solutions and see how they have done things for comparison.

aleksihakli: Other frameworks seem to mutate the request object as well so I'm merging this approach.

In case you have a better implementation just open a PR so we can compare the approaches :)

PR #418, opened 2019-03-09
Improve Python and package versioning and test suite

aleksihakli:


aleksihakli:
Since no one opposes, I'll go ahead and merge this since we're aiming for a good quality future release and 3.6+ is the future with more sites being run on 3.6+ than 3.5.

PR #450, opened 2019-06-12
Callable failure limit

bbayles:
This PR makes the `AXES_FAILURE_LIMIT` setting more flexible, by allowing it to be a callable (that returns an integer), similar to `AXES_USERNAME_CALLABLE`.

This allows for customization of the lockout setting based on the request.


aleksihakli:
Thank you for the clean looking and well formed PR :)

aleksihakli:
Released in version 5.0.7 on PyPI.

PR #452, opened 2019-06-13
Stop showing lockout message when lockout is disabled

### LINKED ISSUE #451: Still get lockout message even with AXES_LOCK_OUT_AT_FAILURE set to False
I have AXES_LOCK_OUT_AT_FAILURE set to False, which does work properly, since I can still log in using valid credentials even when I have passed the failure limit. However, if I have passed the failure limit and enter invalid credentials, then Axes still shows the lockout message (or redirect/template depending on your settings), even though the account isn't actually locked.

If AXES_LOCK_OUT_AT_FAILURE is set to False, then shouldn't Axes also suppress any lockout responses on failed logins?

mogzol:
Resolves #451


aleksihakli:
Released in version 5.0.7 on PyPI.

Issue #432, opened 2019-05-05
Enable django-axes only for staff/admin site


csdenboer:
I would like to only enable django-axes for staff accounts (preferably only admin site). According to the documentation this is currently not possible (I can't find anything about it). Is it true? Thanks!

aleksihakli:
Hey! 

Axes does not yet have flags for only enabling it only for the admin site. It would be quite possible to implement this, and if you would like to have the feature, I would recommend giving it a go yourself and opening a PR. The [`request.path`](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.path) property could be one place for implementing a check in authentication backend layer. 

You can implement this in your own project as well with something like:

```python
import re

from axes.backends import AxesBackend
from axes.helpers import toggleable


class AdminBackend(AxesBackend):
    @toggleable
    def authenticate(self, request, *args, **kwargs):
        if re.match('^/admin', request.path):
            super().authenticate(request, *args, **kwargs)
```

csdenboer:
Hi, thanks for the reply. We have chosen for an alternative solution for now (separate site only running admin). I will submit a PR in case we implement a solution thats worth merging. Thanks for all your efforts.

aleksihakli:
Glad to hear found a solution for your site @csdenboer! 

I would like to leave this issue open, as this is a very good feature request, and others might wish to discuss about and implement this functionality.

aleksihakli:
Resolved by PR #459.

getup8:
@aleksihakli do you mind clarifying what the AXES_ONLY_ADMIN_SITE flag does exactly?  I've set it in my project and failed login attempts on my normal site are still recorded in the database (although I don't think lockout actually occurs).  Is that WAI?  Ideally, I'd like as little axes code as possible to be run and no database calls to be made when someone logs in to the normal site.

hramezani:
@getup8 based on documentation

> AXES_ONLY_ADMIN_SITE : If True, lock is only enable for admin site

It means the locking just works for the admin site and failed login attempts still recorded for all requests.

aleksihakli:
The flag allows login on other URLs than admin. On admin it checks the lockout statuses normally.

https://github.com/jazzband/django-axes/blob/ffc396050b5214ddcd1b55e05a66453c5d1e87ad/axes/handlers/base.py#L153

[...]

PR #459, opened 2019-07-05
Added AXES_ONLY_ADMIN_SITE flag.

hramezani:
[issue 432](https://github.com/jazzband/django-axes/issues/432)

### DIFF
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -20,6 +20,9 @@ class AxesAppConf(AppConf):
     # lock out with username and never the IP or user agent
     ONLY_USER_FAILURES = False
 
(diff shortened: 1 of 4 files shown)


aleksihakli:
Thanks for the PR! Could you please move the implementation of admin checking into the handler? It would be nice to keep the middleware as simple and unopinionated as possible.

hramezani:
@aleksihakli Thanks for the comment. The implementation has moved to the handler.

aleksihakli:
Thanks, the implementation looks very good :) Upstreaming this. I will check the other PRs as well and will aim to make a new release with these changes included soon.

Issue #455, opened 2019-06-19
Add reset command for Access logs


OndrejIT:
Because deleting in admin interface is slow and command can be run by cron..

aleksihakli:
Hey @OndrejIT! That is a good idea, let's add this to feature requests. The feature should be relatively simple in itself. 

I'd imagine that adding a parameter for setting the age would be good since most people probably would like to remove access logs that are e.g. over a month old.

aleksihakli:
Resolved by PR #460 @OndrejIT :)

PR #460, opened 2019-07-05
added delete_access_logs management command

tlebrize:
[#455](https://github.com/jazzband/django-axes/issues/455)
I'm really unsure about the wording in the help and the position/naming of the new test case, feedback would be greatly appreciated.


aleksihakli:
Thank you! I'll change the name to `axes_reset_*` since we have the other commands named similarly in the upstream and try and release this soon.

PR #461, opened 2019-07-08
Don't trigger axes.W003 for subclasses of AxesBackend

adamchainz:
The [usage documentation](https://django-axes.readthedocs.io/en/latest/3_usage.html) advises to create subclass of `AxesBackend` to ignore the lack of `request` if necessary. I've done this in a project using `django-oauth-toolkit`, which doesn't pass `request` (though it should as per [this PR](https://github.com/jazzband/django-oauth-toolkit/pull/643)).

This meant that the axes.W003 check was being triggered, so I've fixed it to check for subclasses of `AxesBackend` as well as the class itself.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,11 @@
 Changes
 =======
 
+Pending
+-------
+
+- Stop axes.W003 check from being triggered for subclasses of ``AxesBackend``.
+  [adamchainz]
 
 5.0.7 (2019-06-14)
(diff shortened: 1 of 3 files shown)


aleksihakli:
Thank you! I'll change the name to `axes_reset_*` since we have the other commands named similarly in the upstream and try and release this soon.

adamchainz:
Thanks!

Issue #476, opened 2019-07-31
BUG: Version 1.6.1: Django 2.2 request.user.is_authenticated is `bool` not function


simanto604newscred:
Python: 3.7.2
Django:2.2
django-axes:1.6.1

in `check_request` function `request.user.is_authenticated` is `bool` not function in Django 2.2.

PR #477, opened 2019-07-31
BUGFIX#476: TypeError for django 2.2 

simanto604newscred:
This needs to be merged under 1.6.1 tag.

### DIFF
--- a/axes/decorators.py
+++ b/axes/decorators.py
@@ -397,6 +397,16 @@ def is_already_locked(request):
     return False
 
 
+def is_authenticated(request):
+    if not hasattr(request, 'user'):
+        return False
+
+    if callable(request.user.is_authenticated):
+            return request.user.is_authenticated()
+
+    return request.user.is_authenticated
+
+
 def check_request(request, login_unsuccessful):
     ip_address = get_ip(request)
     username = request.POST.get(USERNAME_FORM_FIELD, None)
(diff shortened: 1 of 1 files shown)


axes/decorators.py
```diff
@@ -453,8 +453,12 @@ def check_request(request, login_unsuccessful):
     if failures >= FAILURE_LIMIT and LOCK_OUT_AT_FAILURE and user_lockable:
         # We log them out in case they actually managed to enter the correct
         # password
-        if hasattr(request, 'user') and request.user.is_authenticated():
-            logout(request)
+        try:
```
aleksihakli: I think try - catch might not be the easiest way of implementing this check. It could be implemented as a simple function that abstracts away all the logic and is easy to test with different setups:

```python
def is_authenticated(request):
    if not hasattr(request, 'user'):
        return False

    if callable(request.user.is_authenticated):
            return request.user.is_authenticated()

    return request.user.is_authenticated
```

Then it's possible to just use:

```python
if is_authenticated(request):
    logout(request)
```

aleksihakli:
Are you using version 1.6? It's over 2 years old, and we are moving at 5.0 at the moment. It's of course possible to backport this and make a maintenance branch for 1.6 and release 1.6.2 but supporting 1.6 seems odd. Is that surely the case you are thinking about?

simanto604newscred:
> Are you using version 1.6? ... Is that definitely the case you are thinking about?

I am using 1.6.1 since we have a heavy usage of this version on one of our legacy codebase which we can't refactor anytime soon.

axes/decorators.py
simanto604newscred: Changed the PR accordingly. FYI: my changes are on top of 1.6.1 release, is that okay ?

aleksihakli: Yeah it's OK. I'll merge the changes into master and make a maintenance branch for 1.6 and make a release for 1.6.2. It should be a relatively small effort :)

aleksihakli:
Released in [1.6.2](https://pypi.org/project/django-axes/1.6.2/), but please note that you might have [mixed success with recent Django releases](https://travis-ci.org/jazzband/django-axes/builds/567082434).

aleksihakli:
Fixes #476.

Issue #502, opened 2019-10-15
cooloff_time context variable is in iso8601 format


jstockwin:
When using a custom template (by setting the `AXES_LOCKOUT_TEMPLATE`), the template is passed `cooloff_time` as a context variable. This sounds really useful, as I can automatically display to my user "you have been locked out for X minutes/hours".

However, [the code](https://github.com/jazzband/django-axes/blob/93bb73552e522eadd2cc02512183bc05680d43bb/axes/helpers.py#L289) converts this to iso8601 format. In my case, I'm locking users out for 2 minutes, and in iso8601 this reads "You have been locked out for PT2M" - which isn't a very helpful message for the user.

I appreciate you probably want some backwards compatibility here, but what would you say to adding an additional context variable, perhaps `cooloff_timedelta`, which gives me the actual object. This would allow me to add my own formatting. Obviously I could write a template tag that interprets the iso8601 format, but that feels like a bit of a pain since you already have access to the timedelta object internally.

I'm happy to create a PR for this, but wanted to check if/how you'd like this done first.

aleksihakli:
Sounds like a good idea @jstockwin! Please do feel free to open a PR for this.

PR #504, opened 2019-10-16
Return cooloff timedelta as context variable

jstockwin:
Closes https://github.com/jazzband/django-axes/issues/502


aleksihakli:
Upstreamed and released in 5.0.18 on PyPI now.

PR #508, opened 2019-10-24
Added support for DRF json payload in request

### LINKED ISSUE #495: Add Django REST Framework JSON payload support for login
Consider the following curl request:

```
curl --request POST \
  --url http://localhost:8000/login/ \
  --header 'content-type: application/json' \
  --data '{
    "username" : "auser@foo.com",
    "password" : "dshtyh"
}'
```

Django rest framework places the data payload for the above in its `request.data` as opposed to the `request.POST` dictionary

https://www.django-rest-framework.org/tutorial/2-requests-and-responses/#request-objects

As such, any username is not picked up which practically breaks any integration that doesn't use IP or user agent to cross protect because of the following line of code

https://github.com/jazzband/django-axes/blob/master/axes/helpers.py#L145

sk364:
Closes #495

### DIFF
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -93,3 +93,6 @@ class AxesAppConf(AppConf):
         "AXES_META_PRECEDENCE_ORDER",
         getattr(settings, "IPWARE_META_PRECEDENCE_ORDER", ("REMOTE_ADDR",)),
     )
+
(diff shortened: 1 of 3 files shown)


aleksihakli:
Looks good, thanks for the contribution! Adding the flag to docs would be nice as well.

Issue #491, opened 2019-09-17
Optimize access attempt fetching


allixx:
django-axes-5.0.13, django-1.11.23, python 3.6.5

On each login failure multiple (3x) 'Getting access attempts that are newer than' checks are performed.

Are they really neccessary? I didn't look deep enough but it seems one 'access attempt' check after 'expired access attempts' is just enough.

>2019-09-17 07:27:21 [25390] [DEBUG] AXES: Getting access attempts that are newer than 2019-09-17 06:27:21.438521+00:00
>2019-09-17 07:27:21 [25390] [INFO] AXES: Cleaned up 0 expired access attempts from database that were older than 2019-09-17 06:27:21.567521+00:00
>2019-09-17 07:27:21 [25390] [DEBUG] AXES: Getting access attempts that are newer than 2019-09-17 06:27:21.567521+00:00
>2019-09-17 07:27:21 [25390] [WARNING] AXES: Repeated login failure by {username: "user", ip_address: "127.0.0.1", user_agent: "Mozilla/5.0 (Windows NT 6.1; Win64; x64; rv:69.0) Gecko/20100101 Firefox/69.0", path_info: "/admin/login/"}. Count = 2 of 10. Updating existing record in the database.
>2019-09-17 07:27:21 [25390] [DEBUG] AXES: Getting access attempts that are newer than 2019-09-17 06:27:21.567521+00:00
[17/Sep/2019 07:27:21] "POST /admin/login/?next=/admin/ HTTP/1.1" 200 1810

This is a brand new clean django test project. django-axes is basically configured by default, excluding the following:
>AXES_FAILURE_LIMIT = 10
>AXES_RESET_ON_SUCCESS = True

'django.contrib.auth.backends.ModelBackend' auth backend is used.

aleksihakli:
Hey! Good catch. The code is not super optimized. If you'd like, there is room for improvement PRs and database backend improvements.

allixx:
I may have been too quick to judge, axes.attempts.get_user_attempts() is used as a helper in a couple of places to get/update different data.

Still it probably worth to be looked upon by someone more proficient.

aleksihakli:
Leaving this open as there still are multiple points of database access for resets etc even after the good optimizations made by @hramezani.

It would be nice to optimize even further so that this plugin does not cause any unnecessary database load. Multiple queries per requesr can hit a high traffic site pretty hard.

aleksihakli:
[...]

PR #514, opened 2019-11-04
Optimize user attempts fetching. (#491)

hramezani:


aleksihakli:
Looks like a good optimization! It would be nice to check the other database queries and consider if they could benefit from similar optimizations.

aleksihakli:
Released in version 5.0.19 on PyPI.

PR #515, opened 2019-11-05
Added username to attempts update in AxesDatabaseHandler.user_login_failed().

hramezani:
Problem:
In default mode(using just the IP address in get_user_attempts), when login failed, `user_login_failed` add a new record to `AccessAttempt` model.
If login failed again with the same IP address but another username, the record in `AccessAttempt` model still has the old username.

This patch will fix this problem.

### DIFF
--- a/axes/tests/test_handlers.py
+++ b/axes/tests/test_handlers.py
@@ -199,6 +199,32 @@ class AxesDatabaseHandlerTestCase(AxesHandlerBaseTestCase):
         self.assertFalse(AxesProxyHandler().is_locked(self.request, self.credentials))
         self.assertEqual(1, is_whitelisted.call_count)
 
+    def test_user_login_failed_multiple_username(self):
+        tests = [  # (settings, Second login attempt failures_since_start)
+            ({}, 2),
+            ({"AXES_ONLY_USER_FAILURES": True}, 1),
+            ({"AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP": True}, 1),
+            ({"AXES_USE_USER_AGENT": True}, 2),
+        ]
+        for setting_dict, failures_since_start in tests:
+            with self.settings(**setting_dict):
+                with self.subTest(
+                    settings=setting_dict, failures_since_start=failures_since_start
+                ):
+                    self.login(username="admin")
+                    attempts = AccessAttempt.objects.filter(username="admin")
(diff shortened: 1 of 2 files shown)


hramezani:
@aleksihakli, It is ready for review.

axes/tests/test_handlers.py
```diff
@@ -199,6 +199,26 @@ def test_user_whitelisted(self, is_whitelisted):
...
+            {},
+            {"AXES_ONLY_USER_FAILURES": True},
+            {"AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP": True},
+            {"AXES_USE_USER_AGENT": True},
+        ]
+        for setting_dict in settings_list:
+            with self.settings(**setting_dict):
+                self.login(username="admin")
+                self.assertEqual(
+                    1, AccessAttempt.objects.filter(username="admin").count()
```
aleksihakli: Could we check the number of failed attempts here as well from the `failures_since_start` property like so?

```
self.login(...)
attempts = AccessAttempt.objects.filter(...)
self.assertEqual(1, attempts.count())
self.assertEqual(1, attempts.first().failures_since_start)

self.login(...)
attempts = AccessAttempt.objects.filter(...)
self.assertEqual(1, attempts.count())
self.assertEqual(2, attempts.first().failures_since_start)
```

Also, using [`subTest`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) would make the possible test case possible more explicit:

```
for setting_dict in settings_list:
    with self.settings(**setting_dict):
        with self.subTest(settings=setting_dict):
           # ...
```

hramezani:
@aleksihakli Thanks for the review.

> Could we check the number of failed attempts here as well from the failures_since_start property like so?

Yes, good idea. when using `AXES_ONLY_USER_FAILURES` or `AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP`,  we have one record for each and every failed username. So, in the second failed login a new record will create and `failures_since_start` won't be `2`.

Given that, I changed a the test to handle this.
Also, I used `subTest` as you mentioned.

axes/tests/test_handlers.py
hramezani: Done.

aleksihakli:
Released in version 5.0.19 on PyPI.

Issue #528, opened 2019-11-28
Lockouts removed on refresh when deployed, works fine locally


ckcollab:
Thanks so much for django-axes, really appreciate all of the hard work that has gone into making this as easy to use as possible :)

I'm having a bit of trouble getting lockouts to "stick" on our Heroku deployment. We're not using caching, these are our settings:

```
AXES_FAILURE_LIMIT = 8
AXES_LOCKOUT_TEMPLATE = 'lockout_error.json'
AXES_COOLOFF_TIME = 4
AXES_RESET_ON_SUCCESS = True
```

On local development if we clear cache and refresh the login page, the lockout sticks. But, for some reason, on Heroku the lockout does not stick. I imagine it has something to do with how Heroku resolves IPs, but I found in the changelog:

https://django-axes.readthedocs.io/en/latest/10_changelog.html#id37

> _BREAKING CHANGES. AXES_BEHIND_REVERSE_PROXY AXES_REVERSE_PROXY_HEADER AXES_NUM_PROXIES were removed in order to use django-ipware to get the user ip address [camilonova]_

I checked the `django-ipware` project for any complaints behind "Heroku" and didn't find anything, so I'm assuming it _should be_ working. 

----------------

By turning on `AXES_ONLY_USER_FAILURES` I am seemingly able to get around the problem, but would prefer an IP-style-lockout instead of user based.

aleksihakli:
Hi @ckcollab! The lockouts are dependant on Axes being able to resolve the correct end-user IP addresses. It does this by looking at the remote IP address of the client defined in here:

https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-reverse-proxies

And Django contains the headers in the `request.META` object as described here:

https://docs.djangoproject.com/en/dev/ref/request-response/#django.http.HttpRequest.META

The `AXES_META_PRECEDENCE_ORDER` setting defines the order in which to look for remote address in the `request.META` object attributes. The default value is `['REMOTE_ADDR']` so only the default Django remote is used, which works locally. However, you need more configuration for production in order to resolve the remote client IP address correctly.

If you are using the default `REMOTE_ADDR` for `AXES_META_PRECEDENCE_ORDER` then all clients will presumably appear to be coming from e.g. `127.0.0.1` or the Heroku proxy address. When you have the `AXES_RESET_ON_SUCCESS` enabled and locking based on IP configured, you will be resetting all the logins coming from the (same) IP, which is effectively all logins.

[...]

PR #530, opened 2019-11-28
Add special Heroku instructions

ckcollab:
Relates to #528

Not great with wording (or rst).. will update to whatever you all prefer, just want to help the next person with this problem not overlook this :)


aleksihakli:
I added some generic reverse proxy security notes as well. Upstreaming now, thanks for the contribution!

PR #535, opened 2019-12-11
Change `authenticate` to `login`

kebabhans:
### Motivation for change
`django.contrib.auth.authenticate` won't raise the `user_logged_in` signal, if I understand [this](https://docs.djangoproject.com/en/3.0/ref/contrib/auth/#module-django.contrib.auth.signals) correctly. 

### Suggestion
Correct to `login`

### DIFF
--- a/docs/7_architecture.rst
+++ b/docs/7_architecture.rst
@@ -68,7 +68,7 @@ as locked by setting a special attribute into the request.
 The ``AxesMiddleware`` then processes the request, returning
 a lockout response to the user, if the flag has been set.
 
-Axes assumes that the login views either call the ``authenticate`` method
+Axes assumes that the login views either call the ``login`` method
 to log in users or otherwise take care of notifying Axes of authentication
 attempts and failures the same way Django does via authentication signals.
 


hramezani:
I think you misunderstand it. the `authenticate` method of authentication backed is called and [user_login_failed signal will be sent](https://github.com/django/django/blob/8d087f9a179c287e5bb1e24558b097e0a761bc28/django/contrib/auth/__init__.py#L80) when it failed.

aleksihakli:
Hey @kebabhans! 

Django uses the signals with login, logout, and authentication functions. Failures to authenticate produce the failure signals for login while login and logout functions produce the other signals. Both authenticate and login are used in login views.

Authenticate checks the Django credentials with the defined authentication backends. If one of them succeeds, then an user is returned. If one of them raises PermissionDenied or none of them return a user object, an error signal is triggered.

If a user is is returned, it can be logged in with the login function. This triggers the logged in signal.

If a user is logged out, a logged out signal is triggered.

So, both authenticate and login are typically used in a standard Django authentication flow and login view.

aleksihakli:
You could rephrase this as the sentence is indeed illogical: it would be better to talk about authentication than logins :)

PR #550, opened 2020-01-01
Add a tests backend (Fixes #546)

### LINKED ISSUE #546: Should axes_reset with DummyBackend fail silently?
Should axes_reset with DummyBackend fail silently? I think so, as DummyBackend is going to be used for testing ... I don't care if it's not implemented - the implementation should be a null method.

aidanlister:


aleksihakli:
Closes #546, released in 5.2.0.

PR #554, opened 2020-01-06
Fix typos in 4_configuration.rst

pawl-bb:


aleksihakli:
Good catch, thanks @pawl-bb!

PR #557, opened 2020-01-09
Reversion update: handle case when view doesn't have name

mark-mishyn:


aleksihakli:
@mark-mishyn merged, RTD site will update shortly.

PR #574, opened 2020-03-10
docs: Fix simple typo, timdelta -> timedelta

timgates42:
There is a small typo in axes/tests/test_utils.py.

Should read `timedelta` rather than `timdelta`.


aleksihakli:
Good catch!

PR #578, opened 2020-03-22
 Submitting null byte chars causes the user_login_failed to crash 

ddimmich:
Submitting null byte chars in the post data causes the save to database fail with the following error, preventing bad signing attempts from being recorded:

```
/lib/python3.7/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
ValueError: A string literal cannot contain NUL (0x00) characters.
```

I propose replacing these to something more palatable to the django ORM, and allowing axes to continue its work and block users with multiple failed login attempts that contain NUL characters.

To replicate post data with:

```
POST http://localhost:8000/admin/login/
Content-Type: application/x-www-form-urlencoded

csrfmiddlewaretoken%00=123&username=hello&password=somepass!&next=
```

### DIFF
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -94,8 +94,9 @@ class AxesDatabaseHandler(AxesHandler):  # pylint: disable=too-many-locals
             request.axes_path_info,
         )
 
-        get_data = get_query_str(request.GET)
-        post_data = get_query_str(request.POST)
+        # This replaces null byte chars that crash saving failures, meaning an attacker doesn't get locked out.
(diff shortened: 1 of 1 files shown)


aleksihakli:
Thanks @ddimmich, good catch!

ddimmich:
No worries - thank you for merging!  The issue was found by the https://www.intruder.io guys while pentesting.

aleksihakli:
Now released as 5.3.1. Intruder.io looks very nice, it would be good to set up a test site for testing the project with it!

PR #587, opened 2020-05-20
Allow using AXES_ONLY_ADMIN_SITE when admin url isn't registered

igor-shevchenko:
In my project, I have `AXES_ONLY_ADMIN_SITE = True` and there are several urlconfs switched by [django-subdomains](https://github.com/tkaemming/django-subdomains). Only one of them has the path for Django admin. So, `AxesHandler.is_admin` fails for other urlconfs because it can't reverse `"admin:index"`.

This PR wraps `reverse("admin:index")` with `try ... except`, so the function doesn't break when admin url doesn't exist in current urlconf.

### DIFF
--- a/axes/handlers/base.py
+++ b/axes/handlers/base.py
@@ -1,6 +1,7 @@
 import re
 
 from django.urls import reverse
+from django.urls.exceptions import NoReverseMatch
 
 from axes.conf import settings
 from axes.helpers import (
@@ -158,11 +159,11 @@ class AxesHandler:  # pylint: disable=unused-argument
         """
         Checks if the request is for admin site.
         """
-        if (
-            settings.AXES_ONLY_ADMIN_SITE
-            and hasattr(request, "path")
-            and not re.match("^%s" % reverse("admin:index"), request.path)
-        ):
-            return True
+        if settings.AXES_ONLY_ADMIN_SITE and hasattr(request, "path"):
+            try:
+                admin_url = reverse("admin:index")
+            except NoReverseMatch:
+                return True
+            return not re.match("^%s" % admin_url, request.path)
 
         return False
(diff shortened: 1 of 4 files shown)


hramezani:
@igor-shevchenko, You need to add a test for this change and also update the documentation.

igor-shevchenko:
@hramezani thanks! I'll add the test later.

Regarding the documentation: do you have an idea what part should be updated? I can probably change the description of `AXES_ONLY_ADMIN_SITE` in the [list oi settings](https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-project-settings) and mention that it'll work even when admin url doesn't exist. But I think it isn't necessary because this is the behavior a user would expect.

hramezani:
> @hramezani thanks! I'll add the test later.

@igor-shevchenko  Thanks!

> Regarding the documentation: do you have an idea what part should be updated? I can probably change the description of AXES_ONLY_ADMIN_SITE in the list oi settings and mention that it'll work even when admin url doesn't exist.

IMHO, that would be good to mention it.

igor-shevchenko:
I've added the test and updated the docs.

aleksihakli:
The content for this PR is good and implements the necessary changes to fix the functionality when no admin site URL is defined.

Thanks for the review @hramezani and thanks for the fix @igor-shevchenko!

igor-shevchenko:
@aleksihakli thanks for accepting the PR! Are there any plans to release v5.3.3 on PyPI?

aleksihakli:
Hey @igor-shevchenko! 5.3.3 is up on PyPI now, I forgot to make the release before a trip. @camilonova can also make releases. You can see project maintainers that participate in release management at

https://jazzband.co/projects/django-axes

PR #589, opened 2020-05-23
fixes typo in resetting attempts command

sumit4613:
In documentation:
There is a typo in ``python manage.py axes_reset_username [username ...]``.
This will be ``python manage.py axes_reset_user [username ...]``

### DIFF
--- a/docs/3_usage.rst
+++ b/docs/3_usage.rst
@@ -87,7 +87,7 @@ management commands with the Django ``manage.py`` or ``django-admin`` command he
   will reset all lockouts and access records.
 - ``python manage.py axes_reset_ip [ip ...]``
   will clear lockouts and records for the given IP addresses.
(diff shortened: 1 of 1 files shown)


hramezani:
I think this is not a typo. the management command is `axes_reset_username` and `axes_reset_user` is a link for backwards compatibility which is added in 2f319cfff7f944427fd7a82e2c6e3163de585fb9

aleksihakli:
Hey @sumit4613! As @hramezani said `axes_reset_username` is the new name for the command. The old deprecated name was `axes_reset_user` but it was ambiguous as the argument for it was, in fact, username. The command was updated to use a more explicit name that is in line with the argument naming:

https://github.com/jazzband/django-axes/tree/master/axes/management/commands

The old command name was preserved for backwards compatibility so that users relying on the old management commands would not have to change their implementations because library contributors wanted to refactor old code.

PR #605, opened 2020-07-02
Update captcha example

PetrDlouhy:
- Update captcha example to work with current Django
- Add imports
- add `get_ip_address_from_request()` method


PetrDlouhy:
Ah, I finally realised, that `get_ip_address_from_request()` should be in fact `get_client_ip_address()` from `from axes.helpers import get_client_ip_address`.

I hope, that this PR makes the docs less cryptic.

aleksihakli:
Thanks @PetrDlouhy :)

PR #609, opened 2020-07-03
propagate username to the AXES_LOCKOUT_URL redirect

PetrDlouhy:
In order to make possible to remove locks based on username in locked_out view.


aleksihakli:
Needs a style fix to pass the CI pipeline.

Would you like to elaborate how this is useful and maybe add an example to docs so that other people would be aware how they can customize their lockout templates?

aleksihakli:
This commit has already been upstreamed in #610

Issue #412, opened 2019-03-02
Improve base handler class implementations


aleksihakli:
Improve the handler base class and make it an abstract base class or `abc.ABC` with correctly marked `abc.abstractmethods`. 

This requires refactoring the test suite for the handlers.

jorlugaqui:
@aleksihakli I already started working on this.

aleksihakli:
@jorlugaqui awesome! 

I tried to refactor the implementation a bit, but broke some tests in the process and came to the conclusion that refactoring this also requires touching a lot of test code, so I opted to approach the issue again later.

If you can find clean solutions on the handler tests (and maybe login integration test) side of the things as well, that would possibly make things a lot more maintainable. 

#413 is pretty dependent on this issue as well.

jorlugaqui:
@aleksihakli  Ups! it looks like it has been more than one year I have done nothing. My bad, so if you don't mind, I would re-take this work.

aleksihakli:
@jorlugaqui that would be awesome :) Might be good to coordinate with #608 and #613 as well as @PetrDlouhy has been implementing improvements through the stack.

jorlugaqui:
Hi @aleksihakli,

Do you think we can close this as #618 was merged?

PR #618, opened 2020-07-19
Improve handlers orchestration

jorlugaqui:
This PRs tries to implement #412. The idea behind is:

* To have an abstract class (that behaves like an interface) so new handlers would need to implement it.
* To have a base mixin, which all handlers would need to extend in order to be "type" compliant with the Proxy implementation.

In case you agree with the implementation, we may need to wait for #613 (@PetrDlouhy's PR), so I can adjust this PR again.

### DIFF
--- a/axes/handlers/base.py
+++ b/axes/handlers/base.py
@@ -14,7 +14,45 @@ from axes.helpers import (
 )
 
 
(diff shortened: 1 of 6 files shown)


aleksihakli:
Looks nice, good design!

PR #627, opened 2020-08-12
Removed use of Signal()'s providing_args kwarg, as it is deprecated.

coredumperror:
Django 3.1 deprecated the use of the `providing_args` keyword arg to `Signal()`, since it's being removed in 4.0. The argument was always just informational, so removing it has no behavioral effect.

### DIFF
--- a/axes/signals.py
+++ b/axes/signals.py
@@ -17,7 +17,7 @@ from axes.handlers.proxy import AxesProxyHandler
 log = getLogger(settings.AXES_LOGGER)
 
 
-user_locked_out = Signal(providing_args=["request", "username", "ip_address"])
+user_locked_out = Signal()
 
 
 @receiver(user_login_failed)


axes/signals.py
```diff
@@ -17,7 +17,7 @@
 log = getLogger(settings.AXES_LOGGER)
 
 
-user_locked_out = Signal(providing_args=["request", "username", "ip_address"])
+user_locked_out = Signal()
```
GitRon, line 20: You could add a comment so the information from `providing_args` is not lost forever. :)

coredumperror: Done. :)

I would have liked to add a bit about the signal in the docs (it's not mentioned anywhere), but I wasn't sure where it would belong.

GitRon: Nice! But I think a comment should be sufficient for now anyway.

PR #637, opened 2020-09-10
request.is_ajax() is deprecated

smithdc1:
Warning on Travis that `request._is_ajax()` is deprecated and refers to the [release notes](https://docs.djangoproject.com/en/3.1/releases/3.1/#id2)

```
RemovedInDjango40Warning: request.is_ajax() is deprecated. See Django 3.1 release notes for more details about this deprecation.
    if request.is_ajax():
```

I know the docs say something slightly different -- but this change is consistent with the current [code](https://github.com/django/django/blob/83dea65ed649b5e80111017cd5c476c636872b2d/django/http/request.py#L276) and here is a similar change being implemented by a Django fellow.  https://github.com/django-request/django-request/pull/205/files

### DIFF
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -341,7 +341,7 @@ def get_lockout_response(request, credentials: dict = None) -> HttpResponse:
             }
         )
 
-    if request.META.get('x-requested-with') == 'XMLHttpRequest':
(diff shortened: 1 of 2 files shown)


aleksihakli:
Using `request.META` seems like a good approach, good cross referencing on the issue.

PR #640, opened 2020-09-12
Deprecate django-appconf and use local settings for Axes

aleksihakli:
Could we reduce the number of dependencies in the application by deprecating django-appconf since it has created a few incompatibilities in the past and we don't have a massively complex settings configuration?


hramezani (approved):
Nice job 👍

PR #641, opened 2020-09-12
Updated Black version

smithdc1:
Black recently had a release. This bumps the version and applies the formatting changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ jobs:
   allow_failures:
     - env: DJANGO=master
   include:
-    - python: 3.6
+    - python: 3.8
       env: TOXENV=qa
     - stage: deploy
       env:
--- a/axes/management/commands/axes_reset_user.py
+++ b/axes/management/commands/axes_reset_user.py
@@ -1,1 +1,1 @@
-axes_reset_username.py
+axes_reset_username.py  # type: ignore # pylint: disable=all
--- a/axes/tests/test_login.py
+++ b/axes/tests/test_login.py
@@ -490,7 +490,7 @@ class DatabaseLoginTestCase(AxesTestCase):
         AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP=True, AXES_FAILURE_LIMIT=2
     )
     def test_lockout_by_user_and_ip_allows_when_diff_user_same_ip_using_cache_multiple_attempts(
-        self
+        self,
     ):
         # User 1 is locked out from IP 1.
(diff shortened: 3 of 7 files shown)


aleksihakli:
Seems like the PyPy test environment isn't happy with installing the `typed-ast` module that is a dependency for the newer `black`, this has been a problem with the update for a while. One option would be to specify the black installation in the toolchain so that the PyPy environments do not include or use it. It would be nice to keep testing against PyPy for the unit tests, QA can be run in just the `qa` environment :)

smithdc1:
Sounds good to me as I was thinking of something along those lines. Let me
take a look when I get a chance. Also I'll take a look at mypy as that
looks unhappy with the extra blank line black wants to add to one of the
files.

Also -- do you want to take this one step further. That is, install qa
dependencies only in the qa environment. It _should_ speed up the test runs
across the piece?

aleksihakli:
mypy and black cross configuration can be a bit annoying at times.

Yeah, I think QA dependencies should be split into a separate definition. One option would be to have the QA requirements defined in e.g. a separate requirements file under e.g. `requirements/qa.txt` and `requirements/development.txt`? That way we could indeed have tox configure the QA and the test environment separately with different requirements files and avoid quite a bit of boostrapping time in Travis :)

smithdc1:
> mypy and black cross configuration can be a bit annoying at times.

Oh yes.... 😢  The main branch fails locally for me, but passes on Travis. I get it passing locally and now it fails on Travis...

Sorry, if I end up pushing a few commits to see what Travis likes/does not like.

smithdc1:
I think this is now stable. I was struggling to get everything to pass together. A bump of the python version for the QA version seemed to introduce a bit of stability and the tests now pass. 

Hopefully, this is what you had in mind. Appreciate your thoughts. 🙏

.travis.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   allow_failures:
     - env: DJANGO=master
   include:
-    - python: 3.6
+    - python: 3.8
```
aleksihakli, line 19: This is good, although running the QA test matrix against the earliest version provides better indication of support; versions are usually backwards compatible, but not forwards compatible. However, since we only lint and style check with Python 3.8 and earlier versions already have a good test coverage, I think this is quite suited for the project.

aleksihakli (approved):

aleksihakli:
This seems very good! The only thing that some would discuss or argue is the philosophy of implementing QA in a backwards compatible fashion but I think pragmatism is good in this sense. Thank you 👍

PR #649, opened 2020-09-24
Added ppc64le arch support on travis-ci

kishorkunal-raj:
Hi,
I had added ppc64le(Linux on Power) architecture support on travis-ci in the branch and looks like its been successfully added. Also I had excluded the jobs running on python pypy3 version as pypy3 is still not supported on the ppc64le architecture so far. I believe it is ready for the final review and merge. The travis ci build logs can be verified from the link below.

https://travis-ci.com/github/kishorkunal-raj/django-axes/builds/186406415

Please have a look.

Regards,
Kishor Kunal Raj

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,9 @@ env:
   - DJANGO=3.0
   - DJANGO=3.1
   - DJANGO=master
+arch:
+  - amd64
+  - ppc64le
 jobs:
   fast_finish: true
   allow_failures:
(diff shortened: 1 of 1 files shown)


aleksihakli:
We really don't use anything that architecture specific, so is PowerPC support really necessary? Adding an architecture to the build matrix makes builds take longer and basically consumes more resources on Travis end. Would it be enough for you to run PPC CI tests against your own specific application instead of running whole libraries through PPC test matrices?

Issue #634, opened 2020-09-04
Axes settings seems not loaded by django in production


jeremyrich:
Axes is in INSTALLED_APPS and AUTHENTICATION_BACKENDS.
Version : 5.5.0

When trying to run my CI and just test runner I got this :

Here is the traceback : 
Traceback (most recent call last):
  File "manage.py", line 38, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.7/site-packages/django/core/management/__init__.py", line 357, in execute
    django.setup()
  File "/usr/local/lib/python3.7/site-packages/django/__init__.py", line 24, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python3.7/site-packages/django/apps/registry.py", line 122, in populate
    app_config.ready()
  File "/var/www/jung/badoom/core/apps.py", line 20, in ready
    autodiscover_modules("signals")
  File "/usr/local/lib/python3.7/site-packages/django/utils/module_loading.py", line 47, in autodiscover_modules
    import_module('%s.%s' % (app_config.name, module_to_search))
  File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "/usr/local/lib/python3.7/site-packages/ddtrace/internal/import_hooks.py", line 215, in wrapped_find_and_load_unlocked
    return exec_and_call_hooks(module_name, wrapped, args, kwargs)
  File "/usr/local/lib/python3.7/site-packages/ddtrace/internal/import_hooks.py", line 171, in exec_and_call_hooks
    return wrapped(*args, **kwargs)
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/usr/local/lib/python3.7/site-packages/axes/signals.py", line 15, in <module>
    from axes.handlers.proxy import AxesProxyHandler
[...]

PR #651, opened 2020-09-25
moving to axes_settings

jeremyrich:
inspired by django-rest-framework settings and jazzband/django-oauth-toolkit


jeremyrich:
I was to optimistic ... will need to rework the solution :)

aleksihakli:
PR #652 should fix #634 which seems to be caused by premature use of AXES_LOGGER setting :) I'm closing this PR as it has been fixed upstream. Please reopen if you would like to see the settings improved as well!

PR #656, opened 2020-10-03
Add cooloff_timedelta to docs of lockout template

djbrown:
As of #502/#504 the lockout template also receives `cooloff_timedelta` in its context.
This is very useful, but undocumented.
I wish I knew this was in when I read the docs :)

### DIFF
--- a/docs/4_configuration.rst
+++ b/docs/4_configuration.rst
@@ -61,7 +61,7 @@ The following ``settings.py`` options are available for customizing Axes behavio
 * ``AXES_CACHE``: The name of the cache for Axes to use.
   Default: ``'default'``
 * ``AXES_LOCKOUT_TEMPLATE``: If set, specifies a template to render when a
-  user is locked out. Template receives ``cooloff_time`` and ``failure_limit`` as
+  user is locked out. Template receives ``cooloff_timedelta``, ``cooloff_time`` and ``failure_limit`` as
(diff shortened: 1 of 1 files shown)


docs/4_configuration.rst
```diff
@@ -61,7 +61,7 @@ The following ``settings.py`` options are available for customizing Axes behavio
 * ``AXES_CACHE``: The name of the cache for Axes to use.
   Default: ``'default'``
 * ``AXES_LOCKOUT_TEMPLATE``: If set, specifies a template to render when a
-  user is locked out. Template receives ``cooloff_time`` and ``failure_limit`` as
+  user is locked out. Template receives ``cooloff_timedelta``, ``cooloff_time`` and ``failure_limit`` as
```
hramezani, line 64: If I got it right, Based on [this line](https://github.com/jazzband/django-axes/blob/d96a3e46251e69f6c495b8a07bc2afcfd0a19858/axes/helpers.py#L330) It receives ``username`` as well. So we have to include ``username`` too.

hramezani (changes requested):

docs/4_configuration.rst
djbrown: You're right. I also added `username` now

aleksihakli: Looks good now!

Issue #570, opened 2020-03-02
Lockout returns 400 code instead of 403


solartune:
Hi guys,
I’ve tried to use django-axes with django rest framework. For some reason when a user reaches a limit of login attempts he continues receiving 400 code instead of 403. I was able to find a workaround in my view to return 403 but this probably should work out of the box, right?

I followed the documentation when I was installing the package.

After installing I specified a few addition settings:
```python
AXES_COOLOFF_TIME = datetime.timedelta(minutes=env.int("AXES_COOLOFF_TIME", 1))
AXES_FAILURE_LIMIT = env.int("AXES_FAILURE_LIMIT", 2)
AXES_NEVER_LOCKOUT_GET = True
AXES_RESET_ON_SUCCESS = True
```

I use the latest (5.2.2) version of the package.
Django version is 2.2.10
DRF version is 3.9.0

aleksihakli:
Hey @solartune! I think this might have to do with the way DRF handles authentication and it's custom authentication backends. Are you using the defaults with it? Have you checked the integration documentation?

https://django-axes.readthedocs.io/en/latest/6_integration.html#integration-with-django-rest-framework

ckcollab:
That integration should only have to be done for DRF < 3.7 though? I'm experiencing similar problems, the lockout message is never bubbled up to DRF, using base token authentication views.

solartune:
Hi @aleksihakli!
Thank you for your answer!
Yes, as @ckcollab already mentioned, according to the docs it should be the option only for DRF < 3.7 but I'm using DRF 3.9. Do you think we should try something specifically for DRF higher than 3.7?

aleksihakli:
Have you inspected the DRF token authentication and how it interacts with the stock Django authentication stack?

I think that this has more to do with the DRF and Axes not interacting correctly.

Could you post your full DRF, Axes and Django authentication configuration for reference?

ckcollab:
> Have you inspected the DRF token authentication and how it interacts with the stock Django authentication stack?

We haven't dug into DRF yet, we have tried djoser instead of DRF token and experience similar problems though. Will dig into DRF throughout next week.

In the mean time, I believe these are all the relevant pieces?

```python
# settings.py
INSTALLED_APPS = (
    'axes',
)

AUTHENTICATION_BACKENDS = [
    'axes.backends.AxesBackend',
    'django.contrib.auth.backends.ModelBackend',
]

MIDDLEWARE = (
    ...

[...]

PR #659, opened 2020-10-07
Fix integration with drf

solartune:
This PR solves this issue https://github.com/jazzband/django-axes/issues/570

With this change, if a project uses `django rest framework` the package updates the request inside the middleware and then checks whether the limit has been exceeded.


solartune:
The PR is ready to be reviewed.

FYI @aleksihakli

aleksihakli:
LGTM with tests integrated, upstreaming 👍

Issue #597, opened 2020-06-17
AccessAttempt created for username "None"


victorpantoja:
Hey

We have this auth backend:

```python
class GetTokenAuthBackend(ModelBackend):
    """
    Token based authentication, built for the token generated for the REST API
    by the Django REST Framework.

    Expects the auth token in a get variable called "auth_token"
    """

    def authenticate(self, request, username=None, password=None, **kwargs):
        if not request:
            return None

        auth_token = request.GET.get('auth_token', None)

        if auth_token is None:
            return None

        try:
            token = Token.objects.get(key=auth_token)
        except Token.DoesNotExist:
            return None

        return token.user if token.user.is_active else None
```

It can be used like:

```bash
curl -X DELETE \
  'http://localhost:8000/accounts/api/session/?auth_token=<token here>' \
  -H 'Content-Type: application/json' \
  -d '{"ns":"users","action":"get_user","args":[1, 60]}'
```

If for some reason user is not found, `username` will be `None` and a `AccessAttempt` will be created for username `None`.

Based on other parts of the code, shouldn't this case be just ignored and no entry created?

aleksihakli:
The `user_login_failed` signal emitted is handled at the normal handler which saves the AccessAttempt with username `None` to the database indeed:

https://github.com/jazzband/django-axes/blob/d9b4b10a52bb56bfb2cb29c4df21d802af007c49/axes/handlers/database.py#L71

Should this behaviour be adjusted or documentation added on it?

victorpantoja:
That's a good question.

My API is consumed by android and iOS apps and we had a bug on Android that was calling an endpoint with `auth_token=None`. This made all users to be blocked as all of them had "None" as username.

We were using version 4.5.x, that was already saving entries to `AccessAttempt` with username `null` but was not locking those requests. As soon as we deployed the version 5.3.4, all users got locked right away.

Of course, in this case there was a bad implementation of our android client but, does it make sense to store "null" as username on database?

salzamt:
so this issue just appears when IP addresses are ignored, right? 
because blocking triggers on IP+username basis so None and IP unique pair would be just fine because that means that user enumeration is blocked which is a cool feature.
if the IP address is not part of the blocking criteria, the None user should be ignored i think

victorpantoja:
Hey guys! Any thoughts on that?

victorpantoja:
Thanks, @aleksihakli !

aleksihakli:
[...]

PR #670, opened 2020-10-24
Prevent AccessAttempt creation for None username.

hramezani:
Fixes https://github.com/jazzband/django-axes/issues/597 based on this [comment](https://github.com/jazzband/django-axes/issues/597#issuecomment-651623665).


aleksihakli:
Looks good to me!

PR #678, opened 2020-11-12
Return CORS headers to avoid CORS policy error

vladox:
When calling backend running axes from other server or local dev environment over AJAX a CORS policy error is returned.
----
Access to fetch at 'http://0.0.0.0:8005/de/graphql/' from origin 'http://localhost:8000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.


aleksihakli:
Hey @vladox! Would you mind updating the PR so that the tests pass?

axes/helpers.py
```diff
@@ -342,7 +342,13 @@ def get_lockout_response(request, credentials: dict = None) -> HttpResponse:
         )
 
     if request.META.get("HTTP_X_REQUESTED_WITH") == "XMLHttpRequest":
-        return JsonResponse(context, status=status)
+            json_response = JsonResponse(context, status=status)
+            json_response["Access-Control-Allow-Origin"] = settings.AXES_ALLOWED_CORS_ORIGINS
+            json_response["Access-Control-Allow-Methods"] = "POST, OPTIONS"
```
aleksihakli, line 347: These should be parametrisable. Why are POST and OPTIONS allowed specifically?

axes/helpers.py
```diff
@@ -342,7 +342,13 @@ def get_lockout_response(request, credentials: dict = None) -> HttpResponse:
...
+            json_response[
+                "Access-Control-Allow-Headers"
+            ] = "Origin, Content-Type, Accept, Authorization, x-requested-with"
```
aleksihakli, line 350: Why are these specific headers hardcoded for all users?

aleksihakli (changes requested):
A few notes on the CORS headers. I would like to see configurable headers that default to empty set instead of hardcoding the headers globally for all users?

axes/helpers.py
vladox: @aleksihakli kind of the same question above, see previous answer.

axes/helpers.py
vladox: @aleksihakli because those are the ones needed to perform AJAX calls, without them it's not possible they need to be there by default.

vladox: @aleksihakli i've update the description to state that the problem are ajax calls

vladox:
> Hey @vladox! Would you mind updating the PR so that the tests pass?

I'll have a look on it today

axes/helpers.py
aleksihakli: Would HEAD and possibly GET methods also be good to include or are they obsolete in this specific endpoint?

axes/helpers.py
aleksihakli: Roger.

aleksihakli:
We'll only be able to upstream this after the code passes the tests.

axes/helpers.py
vladox: For an AJAX call are not necessary and this is only about those calls
("HTTP_X_REQUESTED_WITH") == "XMLHttpRequest" means that only happens in the case of an AJAX call

vladox (commented):
@aleksihakli I don't know why the test in Python 3.8 is failing it seems to be a problem in the runner not in the code.

aleksihakli:
Hey @vladox! Python 3.8 is failing because QA and code formatting checks are run as part of it: 

https://github.com/jazzband/django-axes/pull/678/checks?check_run_id=1648798212#step:7:21

image

You can run `black -t py36 axes` in your local environment to autoformat the code to be formatting compatible.

aleksihakli (approved):

aleksihakli:
@vladox looks good. We still have some git conflicts, would it be possible to rebase this on top of the current master? After that we can merge and release it.

PR #679, opened 2020-11-14
Add Python 3.5 for Django 2.2 to tox and travic ci.

hramezani:


aleksihakli:
Hey @hramezani!

Although Django 2.2 supports Python 3.5, the Python release itself is EOL and should be deprecated as such. Do you wish to run tests against it? Libraries and CI support is also deprecating so I would personally encourage dropping support, if there is no heavy reason to maintain it.

hramezani:
@aleksihakli Thanks for the clarification. I just added Python 3.5 to complete the CI matrix for Django 2.2 and we can close it.

PR #681, opened 2020-11-18
Disabling middleware when AXES_ENABLED is off

### LINKED ISSUE #680: Middleware still runs with AXES_ENABLED turned off
I'm integrating django-axes into a project where I'd like to disable it in the tests, except where I specifically turn it on.

Unfortunately, if I follow the instructions in the docs, and set `AXES_ENABLED=False` in my test settings, the logic for the middleware still runs during my tests that hit endpoints, many of which don't expect any database access, so they break.

I can conditionally include the middleware based on whether `AXES_ENABLED` is set, but it gets pretty ugly (and I also need to silence the relevant system check, to avoid a spurious warning when I've not included the middleware).

Apologies if I've misunderstood the documentation, but looking at the middleware code, it appears that it doesn't use `toggleable` nor does it check the enabled flag directly.

I'll put up a PR that I think should help.

ashokdelphia:
I'd like to turn django-axes off in tests by setting AXES_ENABLED, but I find that the middleware still runs when I do (in a project that's also using django-rest-framework).

Please let me know if there's a better place to control this, and I can adjust accordingly.

Fixes #680.

### DIFF
--- a/axes/middleware.py
+++ b/axes/middleware.py
@@ -42,19 +42,20 @@ class AxesMiddleware:
     def __call__(self, request):
         response = self.get_response(request)
 
-        if "rest_framework" in settings.INSTALLED_APPS:
-            AxesProxyHandler.update_request(request)
-            username = get_client_username(request)
-            credentials = get_credentials(username)
-            failures_since_start = AxesProxyHandler.get_failures(request, credentials)
-            if (
-                settings.AXES_LOCK_OUT_AT_FAILURE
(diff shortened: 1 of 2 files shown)


nelsonmonteiro:
I had the same problem and I was about to do the same PR.

aleksihakli:
This seems like a good fix, thanks!

ashokdelphia:
> This seems like a good fix, thanks!

Thank you!

aleksihakli:
Released in 5.9.1.

PR #687, opened 2020-12-15
Fix Django master version name

aleksihakli:
Noticed that the Django master version name might have been misspelled in tox setup.


hramezani (approved):

jezdez:
Good catch!

PR #689, opened 2020-12-16
Move DRF integration into signals

### LINKED ISSUE #673: Extra database requests always made with DRF integration
Hi,

I experienced issues with one extra database query in my tests suite after the update to version 5.8.0. The request is:
```sql
SELECT SUM("axes_accessattempt"."failures_since_start") AS "failures_since_start__sum"
FROM "axes_accessattempt"
WHERE ("axes_accessattempt"."ip_address" = '127.0.0.1'::inet AND "axes_accessattempt"."username" IS NULL AND "axes_accessattempt"."attempt_time" >= '2020-11-03T14:32:57.279609+00:00'::timestamptz)
```

I did some digging and it seems to come from the work done in #659 to fix integration with DRF. From what I understand, we normally just track the login attempts when a user fail to login with `user_login_failed`. So, unless we are on a login view, no extra database requests are made. However, with this update, we always do one extra request, even if we are not on a login view (a plain `ListView` in may case) because of the call to `get_failure_limit` which is always done in the middleware.

Is this the desired behavior?

aleksihakli:
Add documentation on how to enable the integration and remove the logic from global middleware.

Personally I'm wondering if configuring signals per-project is more valid than handling the REST Framework specific logic in Axes.

It would be possible to create a signal handler and connecting that in e.g. `axes/signals.py` as well, but I'm not sure if connecting the signals "globally" is such a good idea. It might create hard-to-debug problems for some people.

I've added documentation on how signals can be connected in applications using DRF.

Fixes #673


sobolevn (approved):
Yes, this will work for me! 👍

PR #701, opened 2021-01-06
Prevent cache entry for None username

cabarnes:
Same issue as #597 but for the cache. Consistent fix with #670

### DIFF
--- a/axes/tests/test_handlers.py
+++ b/axes/tests/test_handlers.py
@@ -355,9 +355,11 @@ class AxesCacheHandlerTestCase(AxesHandlerBaseTestCase):
     def test_whitelist(self, log):
         self.check_whitelist(log)
 
-    @patch.object(cache, 'set')
+    @patch.object(cache, "set")
     @patch("axes.handlers.cache.log")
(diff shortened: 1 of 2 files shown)


hramezani (approved):

axes/tests/test_handlers.py
```diff
@@ -354,6 +355,27 @@ def test_empty_request(self, log):
     def test_whitelist(self, log):
         self.check_whitelist(log)
 
+    @patch.object(cache, "set")
+    @patch("axes.handlers.cache.log")
+    def test_user_login_failed_only_user_failures_with_none_username(
+        self, log, cache_set
+    ):
+        with self.settings(**{"AXES_ONLY_USER_FAILURES": True}):
```
aleksihakli, line 363: Just as a note the [`@override_settings`](https://docs.djangoproject.com/en/dev/topics/testing/tools/#django.test.override_settings) decorator is pretty handy and sometimes more readable than using the context manager in test code due to fewer indendations :)

aleksihakli (approved):

aleksihakli:
Looks good, thanks for the contribution!

PR #702, opened 2021-01-06
Use override_settings instead decorator context manager.

hramezani:
Addresses [@aleksihakli comment](https://github.com/jazzband/django-axes/pull/701/files#r552967380)


aleksihakli:
Thanks @hramezani!

PR #703, opened 2021-01-06
Clean up tests

aleksihakli:
A cleaner approach than in #700.

This will produce a `bdist_wheel` distribution without the `axes.tests` submodule as intended. `sdist` distribution will include `test` folder present in the target. 

I'm unsure if this will expose a top-level `test` module to the party installing the package? What is the exact behaviour of `sdist` in the case where we have a top-level folder such as this available?


setup.py
```diff
@@ -38,7 +38,7 @@
     python_requires="~=3.6",
     install_requires=["django>=2.2", "django-ipware>=3,<4"],
     include_package_data=True,
-    packages=find_packages(),
+    packages=find_packages(include=["axes"]),
```
jezdez, line 41: Have you confirmed that this will include the subpackages under `axes/` in the sdist and wheel?

jezdez:
> I'm unsure if this will expose a top-level `test` module to the party installing the package? What is the exact behaviour of `sdist` in the case where we have a top-level folder such as this available?

I think you want to use `find_packages(exclude=["tests", "tests.*"])` for this.

setup.py
aleksihakli: Yes, `python setup.py sdist` and `python setup.py bdist_wheel` do appear to work correctly with this definition and do include e.g. `axes.handlers` and other submodules in the package sources.

aleksihakli:
> > I'm unsure if this will expose a top-level `test` module to the party installing the package? What is the exact behaviour of `sdist` in the case where we have a top-level folder such as this available?
> 
> I think you want to use `find_packages(exclude=["tests", "tests.*"])` for this.

Would `find_packages(include=["axes"], exclude=["tests"])` be the best, most explicit approach here? Trying that, updating the PR for it.

setup.py
aleksihakli: Ended up using `exclude=["tests"]` which seems to do the trick. All other Python modules are included as default.

aleksihakli:
Iterated on the PR and...

- Cleaned up redundant `MANIFEST.in` entries for the `axes` package; source code and localisations are included in the package anyway.
- Moved `axes.tests` to `tests` so it's outside the package, and excluding it from the distribution wheel, reducing package size for users only interested in the runtime package.
- Cleaned up imports and deprecated old `test_utils` module in favour of `test_helpers`, making test code more correlated and readable.

I think it's ready for merging if you folks think it's OK @jezdez and @hramezani.

hramezani (approved):
LGTM :+1: Thanks for the effort!

jezdez:
@aleksihakli Hey, in case you want to worry less about MANIFEST.in, I think it's worth noting the ability of setuptools_scm to act as a replacement for it (mostly).

aleksihakli:
> @aleksihakli Hey, in case you want to worry less about MANIFEST.in, I think it's worth noting the ability of setuptools_scm to act as a replacement for it (mostly).

Oh [those file finders do look handy](https://github.com/pypa/setuptools_scm/#file-finders-hook-makes-most-of-manifestin-unnecessary). I think they are actually tracking all of the SCM files for the sdist as it is. I looks like removing MANIFEST.in altogether produces a correct looking package. I'll go ahead and drop MANIFEST.in from the repository and let's see how the package looks after that. The README and CHANGES files contents are included in the package info anyway and LICENSE is included as part of the stock packaging, so everything should be fine with those.

aleksihakli:
Managed to clean up a bunch of files and moved configuration to `pyproject.toml` in the process, simplifying the main level project structure.

aleksihakli:
@jezdez @hramezani I tagged a release and verified that the archives that were produced by GitHub Actions CI and uploaded to the Jazzband site (both `sdist` gztar and `bdist_wheel` zip archives) looked valid in comparison to the previous release, so the structuring should be good and the wheel should be a bit more lightweight. Thanks for the reviews and tips :)

PR #718, opened 2021-02-18
Fix docs AXES_HANDLER default handler

KStenK:


hramezani (approved):

aleksihakli:
Thank you @KStenK! This will be autoupdated to RTD by the automation soon.

PR #719, opened 2021-02-19
Set AXES_VERBOSE default to AXES_ENABLED

christianbundy:
Problem: When `AXES_ENABLED == False` we still see log output because `AXES_VERBOSE == True`.

Solution: Change `AXES_VERBOSE` default so that if django-axes is disabled then we don't output to stdout.


aleksihakli:
Seems like a valid approach, merging :)

christianbundy:
Thanks for the quick merge!

aleksihakli:
Released in 5.13.1

PR #724, opened 2021-03-01
Mark flaky tests on database handler tests on PyPy runtime

aleksihakli:
The database handler tests fail from time to time due to slow PyPy startup time
which results in tests failing unexpectedly which only happens on PyPy.

This PR requires CPython runtime to pass the tests and indicates failures,
but does not fail the test suite due to them.

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -1,4 +1,7 @@
+from platform import python_implementation
 from unittest.mock import MagicMock, patch
+
+from pytest import mark
 
 from django.core.cache import cache
 from django.test import override_settings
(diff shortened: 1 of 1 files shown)


hramezani (approved):
LGTM :+1:

hramezani:
Ah! I was a little bit late!

aleksihakli:
@hramezani I actually didn't mean to merge this, just to rebase, but thanks for the review. I was wondering on whether or not to fix the test case but I think if it passes on Python 3 it should be good to go since it's verified, PyPy3 just has a bit more latency in driving certain cases and gives false negatives in the test scenario with the database handlers?

PR #733, opened 2021-04-06
Add support for Django 3.2

hramezani:

PR #734, opened 2021-04-06
Drop Django3.0 support.

hramezani:


aleksihakli (approved):
LGTM! This could be released with #733?

PR #737, opened 2021-04-13
Add supported Python and Django badges to README.rst

hramezani:


aleksihakli:
Thanks, seems like a good addition!

PR #738, opened 2021-04-15
Add option to cleanse sensitive GET and POST params in database handler

mcoconnor:
This should allow users to specify which parameters that may be included in GET or POST data should be excluded from the logging

### DIFF
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -118,3 +118,12 @@ settings.AXES_META_PRECEDENCE_ORDER = getattr(
 
 # set CORS allowed origins when calling authentication over ajax
 settings.AXES_ALLOWED_CORS_ORIGINS = getattr(settings, "AXES_ALLOWED_CORS_ORIGINS", "*")
+
+# set the list of sensitive parameters to cleanse from get/post data before logging
+settings.AXES_SENSITIVE_PARAMETERS = getattr(
+    settings,
+    "AXES_SENSITIVE_PARAMETERS",
+    [
+        "password",
+    ],
+)
--- a/axes/conf.py
+++ b/axes/conf.py
@@ -123,7 +123,5 @@ settings.AXES_ALLOWED_CORS_ORIGINS = getattr(settings, "AXES_ALLOWED_CORS_ORIGIN
 settings.AXES_SENSITIVE_PARAMETERS = getattr(
     settings,
     "AXES_SENSITIVE_PARAMETERS",
-    [
-        "password",
-    ],
+    [],
 )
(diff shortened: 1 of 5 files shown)


axes/conf.py
```diff
@@ -118,3 +118,12 @@
 
 # set CORS allowed origins when calling authentication over ajax
 settings.AXES_ALLOWED_CORS_ORIGINS = getattr(settings, "AXES_ALLOWED_CORS_ORIGINS", "*")
+
+# set the list of sensitive parameters to cleanse from get/post data before logging
+settings.AXES_SENSITIVE_PARAMETERS = getattr(
```
aleksihakli, line 123: The docs mention `AXES_SENSITIVE_PARAMS` which differs from this configuration key.

docs/4_configuration.rst
```diff
@@ -111,6 +111,8 @@ The following ``settings.py`` options are available for customizing Axes behavio
   Default: ``False``
 * ``AXES_ALLOWED_CORS_ORIGINS``: Configures lockout response CORS headers for XHR requests.
   Default: ``*``
+* ``AXES_SENSITIVE_PARAMS``: Configures POST and GET parameter values to mask in login attempt logging.
```
aleksihakli, line 114: Configuration key is `settings.AXES_SENSITIVE_PARAMETERS`, should this be the same?

axes/conf.py
```diff
@@ -118,3 +118,12 @@
...
+    settings,
+    "AXES_SENSITIVE_PARAMETERS",
+    [
+        "password",
```
aleksihakli, line 127: Should other sensitive parameters be concatenated to `["password"]` by default? If a user defines a list that overrides this and doesn't have `"password"` included as an element then it will be visible in the logs.

aleksihakli (changes requested):
Looks good. I was wondering if we should somehow offer backwards compatibility for `AXES_PASSWORD_FORM_FIELD` by including that field in the sensitive parameters definition?

aleksihakli:
Hi @mcconnor! Have you had the chance to take a look at the review notes?

mcoconnor:
Thanks very much for the review @aleksihakli .  I think these are good notes and I'd be happy to update--I've been pretty swamped this week but will try to get an update ASAP.

docs/4_configuration.rst
mcoconnor: I've made the naming consistent

axes/conf.py
mcoconnor: I've made the naming consistent

axes/conf.py
mcoconnor: This is a good point.  I've consolidated the handling of ``AXES_SENSITIVE_PARAMETERS``, ``AXES_PASSWORD_FORM_FIELD`` and also hard-coded a case for ``'password'`` to match the existing functionality.

mcoconnor:
Hi @aleksihakli Please let me know if you have any further thoughts on this.  Thanks for your time in reviewing!

aleksihakli (approved):

aleksihakli:
Looks good, thanks for the contribution @mcoconnor! I'll upstream and bake a new release.

PR #755, opened 2021-06-07
fixes default auto field warning

zkanda:
~~[Django will use bigint](https://docs.djangoproject.com/en/3.2/releases/3.2/#customizing-type-of-auto-created-primary-keys) data type in the near future, this PR fixes the warnings:~~

Edit: Turns out we can't use the new bigint because the project still needs to support older version of django, updated the PR to reflect that.

```
WARNINGS:
axes.AccessAttempt: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
axes.AccessLog: (models.W042) Auto-created primary key used when not defining a primary key type, by default 'django.db.models.AutoField'.
        HINT: Configure the DEFAULT_AUTO_FIELD setting or the AppConfig.default_auto_field attribute to point to a subclass of AutoField, e.g. 'django.db.models.BigAutoField'.
```


aleksihakli:
Releasing as part of 5.18.0 :)

PR #763, opened 2021-06-22
Handling multi process applications

uli-klank:
I was confronted with the following crash, while working with a multi process django application:

```bash
Traceback (most recent call last):
   File "path/env/lib/python3.8/site-packages/graphql/execution/executor.py", line 452, in resolve_or_error
     return executor.execute(resolve_fn, source, info, **args)
   File "path/env/lib/python3.8/site-packages/graphql/execution/executors/sync.py", line 16, in execute
     return fn(*args, **kwargs)
   File "path/env/lib/python3.8/site-packages/graphql_jwt/middleware.py", line 73, in resolve
     user = authenticate(request=context, **kwargs)
   File "path/env/lib/python3.8/site-packages/django/contrib/auth/__init__.py", line 84, in authenticate
     user_login_failed.send(sender=__name__, credentials=_clean_credentials(credentials), request=request)
   File "path/env/lib/python3.8/site-packages/django/dispatch/dispatcher.py", line 177, in send
     return [
   File "path/env/lib/python3.8/site-packages/django/dispatch/dispatcher.py", line 178, in <listcomp>
     (receiver, receiver(signal=self, sender=sender, **named))
   File "path/env/lib/python3.8/site-packages/axes/signals.py", line 28, in handle_user_login_failed
     AxesProxyHandler.user_login_failed(*args, **kwargs)
   File "path/env/lib/python3.8/site-packages/axes/helpers.py", line 476, in inner
     return func(*args, **kwargs)
   File "path/env/lib/python3.8/site-packages/axes/handlers/proxy.py", line 100, in user_login_failed
     return cls.get_implementation().user_login_failed(
   File "path/env/lib/python3.8/site-packages/axes/handlers/database.py", line 124, in user_login_failed
     attempt = AccessAttempt.objects.get(
   File "path/env/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
     return getattr(self.get_queryset(), name)(*args, **kwargs)
   File "path/env/lib/python3.8/site-packages/django/db/models/query.py", line 433, in get
        raise self.model.MultipleObjectsReturned(
axes.models.AccessAttempt.MultipleObjectsReturned: get() returned more than one AccessAttempt -- it returned 2!
```
I assume it is created by two issues: the uniqueness of the combination of the user + IP + client is expected, but not enforced by either database nor creation. I adapted the obvious creation in the PR in order to use the database to ensure the atomicity of the get or create, but a solution which also passes the uniqueness requirement to the database would be preferred.
ala:

    class Meta:
       unique_together = ("user", "title")
see https://docs.djangoproject.com/en/3.2/ref/models/options/

### DIFF
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -125,7 +125,14 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
                 username=username,
                 ip_address=request.axes_ip_address,
                 user_agent=request.axes_user_agent,
-                defaults={"failures_since_start": failures_since_start}
+                defaults={
+                    "get_data": Concat("get_data", Value(separator + get_data)),
+                    "post_data": Concat("post_data", Value(separator + post_data)),
+                    "http_accept": request.axes_http_accept,
+                    "path_info":  request.axes_path_info,
+                    "failures_since_start": failures_since_start,
+                    "attempt_time":  request.axes_attempt_time                    
+                }
             )
             # Update failed attempt information but do not touch the username, IP address, or user agent fields,
             # because attackers can request the site with multiple different configurations
@@ -133,14 +140,6 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
 
             separator = "\n---------\n"
 
-            attempt.get_data = Concat("get_data", Value(separator + get_data))
-            attempt.post_data = Concat("post_data", Value(separator + post_data))
-            attempt.http_accept = request.axes_http_accept
-            attempt.path_info = request.axes_path_info
-            if not created:
-                attempt.failures_since_start += 1
-            attempt.attempt_time = request.axes_attempt_time
-            attempt.save()
             # Record failed attempt with all the relevant information.
             # Filtering based on username, IP address and user agent handled elsewhere,
             # and this handler just records the available information for further use.
@@ -150,6 +149,14 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
                     client_str,
                 )
             else:
+                attempt.get_data = Concat("get_data", Value(separator + get_data))
+                attempt.post_data = Concat("post_data", Value(separator + post_data))
+                attempt.http_accept = request.axes_http_accept
+                attempt.path_info = request.axes_path_info
+                attempt.failures_since_start += 1
+                attempt.attempt_time = request.axes_attempt_time
+                attempt.save()
+                
                 log.warning(
                     "AXES: Repeated login failure by %s. Count = %d of %d. Updating existing record in the database.",
                     client_str,
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -119,9 +119,9 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
         # 2. database query: Calculate the current maximum failure number from the existing attempts
         failures_since_start = 1 + self.get_failures(request, credentials)
 
-        # 3. database query: Insert or update access records with the new failure data
-        try:
-            attempt = AccessAttempt.objects.get(
+        if not (settings.AXES_ONLY_USER_FAILURES and username is None):
+            # 3. database query: Insert or update access records with the new failure data
+            attempt, created = AccessAttempt.objects.get_or_create(
                 username=username,
                 ip_address=request.axes_ip_address,
                 user_agent=request.axes_user_agent,
@@ -129,13 +129,6 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
             # Update failed attempt information but do not touch the username, IP address, or user agent fields,
             # because attackers can request the site with multiple different configurations
             # in order to bypass the defense mechanisms that are used by the site.
-
-            log.warning(
-                "AXES: Repeated login failure by %s. Count = %d of %d. Updating existing record in the database.",
-                client_str,
-                attempt.failures_since_start,
-                get_failure_limit(request, credentials),
-            )
 
             separator = "\n---------\n"
 
@@ -146,31 +139,26 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
             attempt.failures_since_start += 1
             attempt.attempt_time = request.axes_attempt_time
             attempt.save()
-        except AccessAttempt.DoesNotExist:
             # Record failed attempt with all the relevant information.
             # Filtering based on username, IP address and user agent handled elsewhere,
             # and this handler just records the available information for further use.
-
-            if not (settings.AXES_ONLY_USER_FAILURES and username is None):
+            if created:
                 log.warning(
                     "AXES: New login failure by %s. Creating new record in the database.",
                     client_str,
                 )
-                AccessAttempt.objects.create(
-                    username=username,
-                    ip_address=request.axes_ip_address,
-                    user_agent=request.axes_user_agent,
-                    get_data=get_data,
-                    post_data=post_data,
-                    http_accept=request.axes_http_accept,
-                    path_info=request.axes_path_info,
-                    failures_since_start=1,
-                    attempt_time=request.axes_attempt_time,
-                )
             else:
                 log.warning(
-                    "AXES: Username is None and AXES_ONLY_USER_FAILURES is enable, New record won't be created."
+                    "AXES: Repeated login failure by %s. Count = %d of %d. Updating existing record in the database.",
+                    client_str,
+                    attempt.failures_since_start,
+                    get_failure_limit(request, credentials),
                 )
+                
+        else:
+            log.warning(
+                "AXES: Username is None and AXES_ONLY_USER_FAILURES is enable, New record won't be created."
+            )
         if (
             settings.AXES_LOCK_OUT_AT_FAILURE
             and failures_since_start >= get_failure_limit(request, credentials)
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -125,6 +125,7 @@ class AxesDatabaseHandler(AbstractAxesHandler, AxesBaseHandler):
                 username=username,
                 ip_address=request.axes_ip_address,
                 user_agent=request.axes_user_agent,
+                defaults={"failures_since_start": failures_since_start}
             )
             # Update failed attempt information but do not touch the username, IP address, or user agent fields,
             # because attackers can request the site with multiple different configurations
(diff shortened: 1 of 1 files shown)


aleksihakli:
To clarify your setup, what transaction isolation level are you running? Are you using atomic transactions on your database engine, and what Django database engine and database versions are you using?

aleksihakli:
@hramezani et al. could you review this PR's code?

axes/handlers/database.py
```diff
@@ -119,58 +119,48 @@ def user_login_failed(
         # 2. database query: Calculate the current maximum failure number from the existing attempts
         failures_since_start = 1 + self.get_failures(request, credentials)
 
-        # 3. database query: Insert or update access records with the new failure data
-        try:
-            attempt = AccessAttempt.objects.get(
+        if not (settings.AXES_ONLY_USER_FAILURES and username is None):
```
aleksihakli, line 122: Doesn't this alter the logic of the code by skipping attempt fetching in case where username is not set? The old logic fetches an access attempt with `username = None` existing in the database and this logic skips that case, right? I'm not sure if that behaviour is of sound design, but this is a change to it nonetheless, and imo should be considered.

Also, I think this logic could be reverted to run the check + logging first so that we immediately return from the function in case the correct conditions are not met, similar to https://github.com/jazzband/django-axes/blob/84b5ccd5d397694d6b04e72e84de9c5861183294/axes/handlers/database.py#L94...L98

axes/handlers/database.py
```diff
@@ -119,58 +119,48 @@ def user_login_failed(
...
+            # 3. database query: Insert or update access records with the new failure data
+            attempt, created = AccessAttempt.objects.get_or_create(
```
aleksihakli, line 124: Using `get_or_create` with `save` will invoke database 3 times in a case where a user does not exist: 

1. first a query to check if it exists, 
2. then a creation if it does not, and
3. finally a save round to update the counters and metadata.

The old logic only uses two database invocations at most, and this MR potentially increases the database query load by 50% per attempt for cases where an access attempt does not previously exist?

- https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create
- https://github.com/django/django/blob/ca9872905559026af82000e46cde6f7dedc897b6/django/db/models/query.py#L571

The proposed logic does avoid a race condition in case where the [database is not configured for atomic requests with `ATOMIC_REQUESTS = True`](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-DATABASE-ATOMIC_REQUESTS) on the engine level AND the unique fields are defined on the model, but this comes at a cost here. Can we define the unique fields in the current configuration, because users can configure the lockout lookup fields with the following settings:

- AXES_ONLY_USER_FAILURES
- AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP
- AXES_LOCK_OUT_BY_USER_OR_IP
- AXES_USE_USER_AGENT

axes/handlers/database.py
```diff
@@ -119,58 +119,48 @@ def user_login_failed(
...
                 username=username,
                 ip_address=request.axes_ip_address,
                 user_agent=request.axes_user_agent,
+                defaults={"failures_since_start": failures_since_start}
```
aleksihakli, line 128: We could use defaults to define the other arguments as well here such as `get_data`, `post_data`, `http_accept`, `path_info`, and `attempt_time` and avoid the roundtrips to database in case the attempt does not previously exist and is being created. If it exists, then we need to fetch it and update the counters and metadata, of course.

axes/handlers/database.py
```diff
@@ -119,58 +119,48 @@ def user_login_failed(
...
-
             separator = "\n---------\n"
 
             attempt.get_data = Concat("get_data", Value(separator + get_data))
             attempt.post_data = Concat("post_data", Value(separator + post_data))
             attempt.http_accept = request.axes_http_accept
             attempt.path_info = request.axes_path_info
-            attempt.failures_since_start += 1
+            if not created:
+                attempt.failures_since_start += 1
```
aleksihakli, line 141: The updates for other counters and metadata should probably be handled in the same logical branch, and `attempt.save()` should only be invoked for requests that exist previously.

axes/handlers/database.py
```diff
@@ -119,58 +119,48 @@ def user_login_failed(
...
             attempt.attempt_time = request.axes_attempt_time
             attempt.save()
```
aleksihakli, line 143: Should this be only invoked only in the case we are updating an existing record? At the moment it does seem to run updates even on newly created items as well, right?

aleksihakli (changes requested):
I think using `get_or_create` with `defaults` dictionary is a sound design but it needs to take into account a few things highlighted in the comments.

axes/handlers/database.py
uli-klank: In the previous implementation, the ```user=None``` case would never have create a database entry, which could have been found in step 3 (prevented due to the if statement in the fallback for new entries), so i needed to pull the condition up to prevent in this case the creation of the entry for no user information.

axes/handlers/database.py
uli-klank: I think the additional save() could be prevented in case of "created"

Marking the whole function as @transaction.atomic might be an easier option, good point 

I do not think that the database configuration has an impact with the current implementation.

axes/handlers/database.py
uli-klank: addressed

axes/handlers/database.py
```diff
@@ -119,58 +119,55 @@ def user_login_failed(
...
             else:
+                attempt.get_data = Concat("get_data", Value(separator + get_data))
+                attempt.post_data = Concat("post_data", Value(separator + post_data))
+                attempt.http_accept = request.axes_http_accept
+                attempt.path_info = request.axes_path_info
+                attempt.failures_since_start += 1
+                attempt.attempt_time = request.axes_attempt_time
+                attempt.save()
+                
                 log.warning(
```
aleksihakli, line 160: Logging is moved from "before the save" to "after the save". Could you please keep the logging for the repeat failure before attempting save? If save fails, no log message is emitted.

axes/handlers/database.py
aleksihakli: Yeah, I agree it didn't do anything sane.

I'd still vouch for reverting this condition to something like:

```python
if settings.AXES_ONLY_USER_FAILURES and username is None:
    log.warning(
        "AXES: Username is None and AXES_ONLY_USER_FAILURES is enabled. New record will NOT be created."
    )
    return
```

axes/handlers/database.py
```diff
@@ -119,58 +119,55 @@ def user_login_failed(
         # 2. database query: Calculate the current maximum failure number from the existing attempts
         failures_since_start = 1 + self.get_failures(request, credentials)
```
aleksihakli, line 120: Looking the code closely this value calculation is not really useful: it is only used in the log message later, and that information is available on the `attempt.failures_since_start` field, I think we should use that instead of this pseudovalue and just mark this whole function with `@transaction.atomic`?

axes/handlers/database.py
```diff
@@ -119,58 +119,55 @@ def user_login_failed(
...
+                    get_failure_limit(request, credentials),
                 )
+                
+        else:
+            log.warning(
+                "AXES: Username is None and AXES_ONLY_USER_FAILURES is enable, New record won't be created."
+            )
         if (
             settings.AXES_LOCK_OUT_AT_FAILURE
             and failures_since_start >= get_failure_limit(request, credentials)
```
aleksihakli, line 173: I think we should just use `attempt.failures_since_start` here after adjusting the conditions above a bit?

axes/handlers/database.py
```diff
@@ -119,58 +119,55 @@ def user_login_failed(
...
             # in order to bypass the defense mechanisms that are used by the site.
 
-            log.warning(
-                "AXES: Repeated login failure by %s. Count = %d of %d. Updating existing record in the database.",
-                client_str,
-                attempt.failures_since_start,
-                get_failure_limit(request, credentials),
-            )
-
             separator = "\n---------\n"
```
aleksihakli, line 141: This separator can be moved inside the `if created` block as well.

aleksihakli (changes requested):

axes/handlers/database.py
uli-klank: The query get_failures is responsible to find similar attempts (e.g. same ip different user), so needs to return to the condition. Sorry my bad, I did not setup the tests for me locally in time.

axes/handlers/database.py
uli-klank: This is still the main check to lock out a user, which did also before not depend on the new entry.

axes/handlers/database.py
uli-klank: The transaction atomic should not be required anymore, and now also understanding the rather complex query inside get_failures, I would rather stick to the local solution.

axes/handlers/database.py
aleksihakli: Ah, true that. So, let's keep it in place for the time being.

axes/handlers/database.py
aleksihakli: OK 👍

aleksihakli (approved):

uli-klank:
Thank you for your support.

hramezani:
Sorry for being late.
Thanks @uli-klank and @aleksihakli  for your great work.

hramezani (commented):
LGTM

aleksihakli:
Thank you for checking out the changes @hramezani 👍

stephenmullens:
> ```shell
> graphql_jwt
> ```

I noticed that you are using graphql_jwt library with django axes. Do you have suggestions or ideas how best I can integrate django axes with django-graphql-jwt? Critically how to integrate it with TokenAuth?

`token_auth = graphql_jwt.ObtainJSONWebToken.Field()`

uli-klank:
The token acquisition in graphql-jwt is providing an authentication backend.
The requests go through the middlewares, including axes (we have axes here as last entry) and through the authentication backend (here we have axes before graphql_jwt). Does this answer the question?

stephenmullens:
> The token acquisition in graphql-jwt is providing an authentication backend. The requests go through the middlewares, including axes (we have axes here as last entry) and through the authentication backend (here we have axes before graphql_jwt). @stephenmullens Does this answer the question?

Thanks for your fast response. I have got it working by the addition of a lockout view:

`AXES_LOCKOUT_CALLABLE = "users.views.lockout"`

`def lockout(*args, **kwargs):
    """
    Generates a simulated graphql response to match the expected syntax in the UI
    reference app django-axes

    args received:
    {'username': 'email@email.com', 'password': '********************'}
    """
    response_obj = {}
    response_obj["errors"] = [{"message" : "Incorrect credentials. Account locked for " + str(settings.AXES_COOLOFF_TIME_MINUTES) + " minutes",
        "path":["tokenAuth"]}
    ]
    response_obj["data"] = {"tokenAuth":None}

    return JsonResponse(response_obj, safe = False, status=200)`

If i don't include the axes middleware, then my auth api still responds properly with a JWT, but that JWT pair doesn't work. Do you have an idea about what might be going on?

uli-klank:
Ah I see, did not mention it: we use the AXES_COOLOFF_MESSAGE with a similar GraphQL statement.

PR #765, opened 2021-06-29
Adjust commentation and log messages

aleksihakli:
@uli-klank made a few adjustments to documentation and logs, do you think these look OK? 

I also moved the `failures_since_start` fetching from top of the handler to the last step, since it should include the necessary information of the added attempts there, is that correct?


uli-klank:
Sounds more logical to me as well.

Issue #728, opened 2021-03-16
Would HTTP 429s be more appropriate?


anthonyryan1:
I recently had a penetration tester question whether or not django-axes was working because it returned the same 403 code as an invalid password attempt.

I did verify it was working correctly, but I thought I would raise the issue for discussion on whether there would be merit to using a 429 status code instead of a 403.

The 429 is arguably more semantically correct than the 403, and would make it easier to distinguish a rejected password vs a blocked attempt via HTTP headers alone.

I look forward to your thoughts. If you feel this is a worthwhile change I can prepare a pull request.

aleksihakli:
HTTP 429 might be semantically more correct:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429

The response could also in principle include the standardized `Retry-After` header to indicate how long the lockout period is.

This would change the current behaviour and be a good candidate for the version 6 release of Axes?

aleksihakli:
@anthonyryan1 this has now been made possible by #770 but I'm wondering if the default could still perhaps be updated from 403 to 429, and would this warrant a major release 6.x as part of the update?

PR #770, opened 2021-08-17
Add ability to change status code

phil-bell:
Am adding the ability to change the http status code returned when the retry limit is reached. Have left the default as `403` and added a test and documentation for the change.

Use case would be something like wanting to return `429` instead.


aleksihakli:
Hey @phil-bell and thanks for the MR! There is some discussion on #728 about the most suitable HTTP status codes as well. I think this MR is good as-is, but do you have an opinion on whether we should use 429 or 403 as the default? If you want to chime in, the issue would be the best place for further discussion as this MR is now merged.

PR #775, opened 2021-08-25
Make failures_since_start accessible to the views

okapies:
Set `failures_since_start` calculated in the `AxesHandler`s to `request` to make the number of failed login attempts visible to the Django views.

### DIFF
--- a/axes/handlers/proxy.py
+++ b/axes/handlers/proxy.py
@@ -76,6 +76,7 @@ class AxesProxyHandler(AbstractAxesHandler, AxesBaseHandler):
             request.axes_user_agent = get_client_user_agent(request)
             request.axes_path_info = get_client_path_info(request)
             request.axes_http_accept = get_client_http_accept(request)
+            request.axes_failures_since_start = 0
             request.axes_updated = True
 
     @classmethod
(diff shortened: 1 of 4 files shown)


axes/handlers/proxy.py
```diff
@@ -76,6 +76,7 @@ def update_request(request):
             request.axes_user_agent = get_client_user_agent(request)
             request.axes_path_info = get_client_path_info(request)
             request.axes_http_accept = get_client_http_accept(request)
+            request.axes_failures_since_start = 0
```
aleksihakli, line 79: This seems like an antipattern to set the value to 0? I'd much rather set it to None if it's not known, than set it to a value that can be misinterpreted?

aleksihakli (changes requested):
The idea seems good, but if we set `request.axes_failures_since_start = 0` in the proxy handler then that opens the door for introducing easy-to-write-hard-to-debug bugs later on, if someone forgets to update that value in their handler?

axes/handlers/proxy.py
okapies: Makes sense. Fixed.

aleksihakli (approved):

aleksihakli:
Hey @okapies! This has now been merged and released as part of version 5.22.0.

PR #777, opened 2021-08-25
Increment failures_since_start in an atomic manner

okapies:
Increment `AccessAttempt.failures_since_start` in the DB instead of the application to make the operation atomic.

### Note
The same issue still remains on `AxesCacheHandler`:

https://github.com/jazzband/django-axes/blob/ac86d4b2139f632cee972e874984982a9628ab27/axes/handlers/cache.py#L100

### DIFF
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -1,6 +1,6 @@
 from logging import getLogger
 
-from django.db.models import Sum, Value, Q
+from django.db.models import F, Sum, Value, Q
 from django.db.models.functions import Concat
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


axes/handlers/database.py
```diff
@@ -152,18 +152,21 @@ def user_login_failed(
...
-                attempt.failures_since_start += 1
+                attempt.failures_since_start = F("failures_since_start") + 1
                 attempt.attempt_time = request.axes_attempt_time
                 attempt.save()
 
                 log.warning(
                     "AXES: Repeated login failure by %s. Count = %d of %d. Updated existing record in the database.",
                     client_str,
-                    attempt.failures_since_start,
+                    # can be different from the actual value in the DB
```
aleksihakli, line 168: This breaks the idea of atomicity, really. Would it be more useful to just deprecate the count from this log message, if we can't reliably present it?

okapies: Yes, I think that's an option. I'll remove it.

aleksihakli:
Hey @okapies! This has now been merged and released as part of version 5.22.0.

Issue #788, opened 2021-09-13
AccessAttempt MultipleObjectsReturned


PetrDlouhy:
I have sometimes following error from my production server:

```

AccessAttempt.MultipleObjectsReturned: get() returned more than one AccessAttempt -- it returned 2!
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 562, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "django/views/decorators/debug.py", line 89, in sensitive_post_parameters_wrapper
    return view(request, *args, **kwargs)
  File "django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "django/contrib/auth/views.py", line 63, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "newrelic/hooks/framework_django.py", line 952, in wrapper
    return wrapped(*args, **kwargs)
  File "django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "django/views/generic/edit.py", line 141, in post
    if form.is_valid():
  File "django/forms/forms.py", line 177, in is_valid
    return self.is_bound and not self.errors
  File "django/forms/forms.py", line 172, in errors
    self.full_clean()
  File "django/forms/forms.py", line 375, in full_clean
    self._clean_form()
  File "django/forms/forms.py", line 402, in _clean_form
    cleaned_data = self.clean()
  File "django/contrib/auth/forms.py", line 215, in clean
    self.user_cache = authenticate(self.request, username=username, password=password)
  File "django/contrib/auth/__init__.py", line 84, in authenticate
    user_login_failed.send(sender=__name__, credentials=_clean_credentials(credentials), request=request)
[...]

PR #789, opened 2021-09-13
fix duplicated AccessAttempts

PetrDlouhy:
fix for #788 

This prevents duplicated AccessAttempt objects to be created, also it deletes those duplicated attempts that are already present during the migration.


aleksihakli:
@camilonova @hramezani @okapies what do you think of the update for the uniqueness constraints here? Can the uniqueness break something? I'd like at least a few people to pitch their opinions in for this change.

aleksihakli:
@PetrDlouhy the QA step is failing with the current configuration and needs to be fixed before upstreaming.

okapies:
I forgot the uniqueness setting… Thanks for the fix!

aleksihakli:
Still needs stylistic fixes for the style checker to pass:

image

aleksihakli (approved):
LGTM!

Issue #750, opened 2021-05-29
Sample AXES_USERNAME_CALLABLE code fails on Lockout


dwasyl:
I'm not sure if this is related to #736, but when implementing `django-axes` with a custom `AXES_USERNAME_CALLABLE` I kept stumbling across a problem on lockout.

Based on the documentation, I created a `get_username` which adds a tenant suffix to the username:
```
def get_username(request, credentials):
    """ Callable to generate the proper username for django-axes """

    username = credentials.get('username')
    namespace = "#%s" % (request.tenant.id) if request.tenant else ""
    return "%s%s" % (username, namespace)
```

This seems to work fine, the access log get entries with the correct suffixed username, but when an account gets locked out,. Instead of the lockout message, I get an error `'NoneType' object has no attribute 'get'`, that stems from https://github.com/jazzband/django-axes/blob/master/axes/middleware.py#L40 calling `response = get_lockout_response(request)` which carries on to https://github.com/jazzband/django-axes/blob/master/axes/helpers.py#L368 calling `"username": get_client_username(request, credentials) or "",` then `get_client_username` eventually calls my original `get_username` except `credential=None`, so the `credentials.get("username")` fails.

Perhaps the sample code in https://django-axes.readthedocs.io/en/latest/5_customization.html#customizing-username-lookups should be updated to check for a valid credentials, or perhaps `get_client_username` shouldn't be called on a locked out account? It would be good to have `username` in the context variables for the file displayed on lockout.

I don't have enough familiarity with the package to identify the best fix for this though.

aleksihakli:
Hi @dwasyl, and thanks for reporting this! Seems like not all code paths support the custom usernames and some of them should be updated and a test added to check this bug and catch a possible regression. Anyone interested in doing this?

aleksihakli:
This should be fixed by #799 and fix is released in version 5.26.0.

PR #799, opened 2021-10-07
#750, fix lockout failure when providing AXES_USERNAME_CALLABLE

rootart:
@aleksihakli  would be great to hear your feedback if it could be an acceptable solution for issue #750. Many thanks in advance.


aleksihakli:
Hey @rootart and thanks for the PR! The approach does seem good. I'll try and find time for reviewing the code and checking the use and flow of credentials in handlers and middleware, but my calendar is a bit swamped so this might take a while.

PR #803, opened 2021-10-21
Add explicit dependecy on setuptools

asherf:
this package depends on [setuptools](https://pypi.org/project/setuptools/) since it [uses pkg_resources](https://github.com/jazzband/django-axes/blob/master/axes/__init__.py#L1)  which is [part of setuptools](https://github.com/pypa/setuptools/tree/main/pkg_resources)
However, that dependecy is not declared which in some uses cases can can an import error when django-axes is being used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,7 +36,7 @@ setup(
     use_scm_version=True,
     setup_requires=["setuptools_scm"],
     python_requires="~=3.6",
-    install_requires=["django>=2.2", "django-ipware>=3,<5"],
+    install_requires=["django>=2.2", "django-ipware>=3,<5", "setuptools"],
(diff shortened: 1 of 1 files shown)


asherf:
@aleksihakli any thoughts on my reply?
Another option I thought about was making a conditional dependency on setuptools for phyton versions that had don't have the new api..
But that will introduce complexity that I am not sure is justified considering in most use cases setuptools is already available (I just noticed this issue in my env because we use a tool that only includes the explicitly required dependencies)

aleksihakli:
Good point, let's go with setuptools then :)

aleksihakli:
Released in version 5.27.0.

Issue #814, opened 2021-11-24
Misleading documentation on django rest framework integration


xi:
This is about this section in the docs:

https://django-axes.readthedocs.io/en/latest/6_integration.html#integration-with-django-rest-framework

The `user_locked_out` signal is triggered in reaction to the stock django `user_login_failed` signal. That signal is
in turn triggered by django's `authenticate()` function which is used to check combinations of username and password.

For API requests to django rest framework you would use one of the [pluggable authentication schemes](https://www.django-rest-framework.org/api-guide/authentication/). Only one of them (`BasicAuthentication`) uses `authenticate()`.

I would expect a proper integration to block brute-force attempts on e.g. `TokenAuthentication`. Either the example from the documentation should be extended to cover such common cases, or there should be a warning that explains these limitations.

aleksihakli:
Hi @xi :) Would you be interested in opening a PR for matching the facts and language of the documentation page to reality?

The documentation page can be edited easily with GitHub editor:

https://github.com/jazzband/django-axes/blob/master/docs/6_integration.rst

PR #834, opened 2022-01-18
warn about limited django rest framework integration

xi:
ref #814


aleksihakli:
Thanks @xi! The documentation should be updated by RTD shortly.

PR #838, opened 2022-01-21
Exclude running test for Python3.7 and Django main.

hramezani:


aleksihakli:
Thank you @hramezani :)

PR #846, opened 2022-02-21
Some tests are skipped due to duplicate names fix

### LINKED ISSUE #845: Some tests are skipped due to duplicate names
Test names must be unique per scope otherwise the second test overrides the first one with the same name.

For example if you had a test file that does:

```
def test_a():
    pass

def test_a():
    pass
```

Then only the second `test_a` will be ran. More details [here](https://codereview.doctor/features/python/best-practice/avoid-duplicate-unit-test-names).

These are the tests that are overriding previously defined tests due to this problem:

https://github.com/jazzband/django-axes/blob/9d391841792e817f38a6924320cb2d0439656153/tests/test_helpers.py#L262

I found this issue automatically, see other issues [here](https://codereview.doctor/jazzband/django-axes)

code-review-doctor:
Fixes #845.

These tests were overriding other tests in the same file with the same name. I renamed so the tests run. Please give the tests a better name than I used. Naming is hard and I'm just a bot :)

This might result in tests failing because the affected tests were previously not running.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -259,7 +259,7 @@ class ClientStringTestCase(AxesTestCase):
     @override_settings(
         AXES_CLIENT_STR_CALLABLE="tests.test_helpers.get_dummy_client_str_using_request"
     )
-    def test_get_client_str_callable(self):
+    def test_get_client_str_callable_two(self):
         self.request.user = self.user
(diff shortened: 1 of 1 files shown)


hramezani:
Superseded by https://github.com/jazzband/django-axes/pull/848

PR #848, opened 2022-02-21
Fix #845 -- Rename test functions

### LINKED ISSUE #845: Some tests are skipped due to duplicate names

hramezani:
Fixes https://github.com/jazzband/django-axes/issues/845

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -248,7 +248,7 @@ class ClientStringTestCase(AxesTestCase):
     @override_settings(
         AXES_CLIENT_STR_CALLABLE="tests.test_helpers.get_dummy_client_str"
     )
-    def test_get_client_str_callable(self):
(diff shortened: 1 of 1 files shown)


aleksihakli:
Thanks for the fix @hramezani :)

PR #855, opened 2022-03-04
Pin pypy-3.8 version to fix SQLite error in CI.

hramezani:
See CI fail here https://github.com/jazzband/django-axes/runs/5421924209?check_suite_focus=true


aleksihakli (approved):
Looks great, thanks!

PR #856, opened 2022-03-07
Feature/cleanup management command

GitRon:
Creating a management command which can be used to purge all attempts and logs from the database being older than X days.


GitRon:
Oops, this feature was already implemented 🙈

PR #862, opened 2022-03-15
Failures logs

p1-gdd:
# Issue
The cool off time is normally used to lockout the user. But there is an unattended behaviour, it is also used to clean expired attempts (attempts that are older than the cool off time threshold).

# Needs
But how is our use case:
* Users MUST be locked-out during a grace period
* Users MUST see a log with every attempts to be informed of attack attempts
* Attempts database MUST never be cleaned automatically
* Attempts MUST be manually cleaned by an administrative human

# Fix-up with this PR
This PR add a new class model named `AccessFailureLog` that record every failed logins, like `AccessLog` do with successful logins.


p1-gdd:
You're right, this was not covered. I complete with your recommendations and push.

p1-gdd:
I think this is a begin. Users will probably find some new kind of use case to improve this PR.

I've base our branch on version 5.28.0 as we DO still use Python 3.6 branch, even if this PR is master based.

aleksihakli (approved):

aleksihakli:
Thank you for the contribution and PR @p1-gdd and sorry for the late review.

aleksihakli:
Released in version 5.32.0.

PR #865, opened 2022-03-22
Fix #837 - Don't reset cooloff time after login attempt during lockout

### LINKED ISSUE #837: Failed logins and attempt time updated even when lockedout
I'm not sure if this is intended or not, didn't find any information in the docs related to this.

The problem is that is an user is lockedout, the failed logins and the attempt time gets updated. This way `AXES_COOLOFF_TIME` is not working as expected, or at least as I would expect.

Example with `AXES_COOLOFF_TIME=1` (hour):
- Current behaviour
1. User fails to login 3 times, the last attempt at 10:05.
2. User gets lockedout until 11:05.
3. User tries to login at 10:50, and failed logins updates to 4, and attempt time updates to 10:50, so the user will be lockedout until 11:50. Also this is true even if the 4th attempt was with the right credentials.

- Expected behaviour
1. User fails to login 3 times, the last attempt at 10:05.
2. User gets lockedout until 11:05.
3. User tries to login at 10:50, the user will not login to the site even with the right credentials, but the attempt will NOT be taken into account.

I think this should be the typical behaviour, because the user probably won't know the exact time of the last failed login, maybe he thinks that the last attempt was at 10:00, and he tries to access at 11:01, but then he will see that now he is going to be lockedout 1 more hour, even if he tried with the right credentials. The current behaviour leads us to a scenario where an user could be lockedout easily for 2 or 3 hours, if he tries to access before the correct time, instead of the intended `AXES_COOLOFF_TIME` of 1 hour.

I hope I explained correctly. Thanks for your time!

antoine-42:
Fixes https://github.com/jazzband/django-axes/issues/837

Adds the `AXES_RESET_COOL_OFF_ON_FAILURE_DURING_LOCKOUT` option to not reset cooloff time in case of a login attempt during lockout. It's set to `True` by default, so the default behavior won't change.

### DIFF
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.contrib.auth.backends import ModelBackend
 
 from axes.exceptions import (
@@ -52,7 +53,8 @@ class AxesBackend(ModelBackend):
         response_context["error"] = error_msg
 
         # This flag can be used later to check if it was Axes that denied the login attempt.
-        request.axes_locked_out = True
+        if not settings.AXES_RESET_COOL_OFF_ON_FAILURE_DURING_LOCKOUT:
+            request.axes_locked_out = True
 
         # Raise an error that stops the authentication flows at django.contrib.auth.authenticate.
         # This error stops bubbling up at the authenticate call which catches backend PermissionDenied errors.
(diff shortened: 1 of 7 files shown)


igonro:
Nice job! LGTM. Let see if @aleksihakli or someone else can review it

aleksihakli (approved):

aleksihakli:
Thank you for the new feature @antoine-42 and sorry for the late review.

aleksihakli:
Released in version 5.32.0.

PR #889, opened 2022-05-15
Fix QA and test issues

aleksihakli:
Continue using MD5 hashing in Axes test settings
as it offers better performance for test runs
without compromising security for users

Fixes issues raised by #888

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,7 +17,7 @@ jobs:
         django-version: ['3.2', '4.0', 'main']
         include:
           # Tox configuration for QA environment
-          - python-version: '3.8'
+          - python-version: '3.10'
             django-version: 'qa'
           # Django main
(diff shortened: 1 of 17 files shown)


hramezani (approved):

.github/workflows/test.yml
```diff
@@ -13,11 +13,11 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', 'pypy-3.8-v7.3.7']
-        django-version: ['3.2', '4.0', 'main']
```
hramezani, line 17: Don't we need django main in CI?

aleksihakli: Django main is specified in the `include` part of the build definition matrix so this declaration has been a duplicate :)

PR #898, opened 2022-05-30
#897: Settings variables as table instead of list in the docs

### LINKED ISSUE #897: Optimise settings variable overview in the docs
I find it kind of hard to read the [docs](https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-project-settings) part of the settings variables. Maybe a table layout helps? I am providing a PR for a start but I can't see how this would look at RTD, though.

Help appreciated 🙂

GitRon:
fixes #897


aleksihakli:
The diff and updating the docs is a bit cumbersome if we always have to use the exact right amount of dashes in the .rST files. Is there an "easier-to-update" way of providing a table layout? I think the readability would probably be better with the table!

GitRon:
@aleksihakli Sorry, didnt get your point. Are you worried that the table might be broken?

aleksihakli:
I am a bit concerned that updating the content from a list to a table might prove to be a bit complex in the long run. The code and git diffs will be easy-to-read-hard-to-update for a table, if we change the format from the current easy-to-update-hard-to-read list.

I think the proposed table could prove to be be better for the users of the documentation, but am trying to take the maintainers into account as well.

I'm not sure if this is a real concern though, so if you think the table is better we can give it a try?

GitRon:
Actually, I have no strong feeling about the table. I just realised that the current version of the docs is really hard to read and to quickly get the required information. I'm open to suggestions 🙂

aleksihakli:
Hey @GitRon this is how it currently looks with the table:

https://django-axes.readthedocs.io/en/latest/4_configuration.html

image

I wonder if there is an option of widening the text element for better readability with the table? Otherwise browsing the documentation might prove to be a bit challenging with the styling.

GitRon:
I think it's an improvement, just the asterisk is shown in a wrong way...

aleksihakli:
The table is a bit challenging to slide sideways to show the description text per configuration variable at the moment, did you try the RTD site out?

GitRon:
Yeah, I struggle with this quite often myself. But if you have only that much space, what can we do? I think, after all, it is still more readable than before and the defaults are more explicit.

Issue #812, opened 2021-11-19
AxesBackend should not handle permission checks


jcgiuffrida:
### Expected behavior
AxesBackend raises `PermissionDenied` if a user is locked out. Otherwise, the other `AUTHENTICATION_BACKENDS` do the rest of the authentication work.

### Actual behavior
AxesBackend handles permissions checks just like Django's `contrib.auth.backends.ModelBackend` because it inherits from it. 

### Explanation
My site uses a custom authentication backend that handles directly-assigned user permissions differently (depending on certain conditions) than Django's contrib.auth.backends.ModelBackend. It took me a while to figure out that AxesBackend was actually messing with this, because it subclasses from ModelBackend, so that it (silently) runs all the methods of ModelBackend except for the `authenticate` method that it overrides. 

For example, suppose that we don't want to use Django's default logic of having superusers inherit all permissions. This is done in ModelBackend._get_permissions. Even if we override that or change it, AxesBackend still gives superusers all permissions when it runs that method. 

I do not believe this is expected or desired behavior, for the small subset of users like me who have a different setup and now have to override AxesBackend to change its inheritance or null out the permission-related methods. In my understanding, Axes is about authentication, not permissions, and does not even need to subclass from ModelBackend at all (as long as it has the [required backend methods](https://docs.djangoproject.com/en/3.2/topics/auth/customizing/#writing-an-authentication-backend)). 

### Proposed fix
I think it should inherit from `object` instead, and we should add a blank `get_user` method that returns `None` to conform with Django's requirements for custom backends. No change to the `authenticate` method, so no change to the Axes part of AxesBackend. However, AxesBackend would no longer run permissions checks because it would no longer inherit from ModelBackend. 

I'm happy to submit a PR to do this, but wanted to check as this is pretty sensitive stuff.

* Is my interpretation correct?
* The [documentation](https://django-axes.readthedocs.io/en/latest/2_installation.html) specifies that ModelBackend should come after AxesBackend, so the majority of users are probably doing that and would not be affected by this. What about users who do **not** currently use ModelBackend - perhaps because they have custom authentication logic in a separate backend - and are relying on AxesBackend to perform permissions checks? They may not even realize it. Perhaps instead of changing AxesBackend we could add a new backend class that inherits from `object` and leave the current one alone - and update the documentation with a note for users with custom authentication backends.

aleksihakli:
Hi @lucaluca, and thanks for submitting the detailed issue description.

[...]

PR #906, opened 2022-07-13
Provide new backend class that does not inherit from ModelBackend

jcgiuffrida:
Fixes https://github.com/jazzband/django-axes/issues/812. 

This PR provides a new backend class called `AxesStandaloneBackend` which inherits from Django's `BaseBackend`, not `ModelBackend`. Therefore, this backend class does not have the rare, but unexpected side effects that `AxesBackend` has for projects that do not implement `ModelBackend`, such as automatically assigning all permissions to superusers. Instead, this backend handles Axes' lockout logic, and nothing else, as advertised.

`AxesBackend` remains unchanged, and its inheritance is made clearer by inheriting from both `AxesStandaloneBackend` and `ModelBackend`. This PR is backwards-compatible and should have no impact on any existing projects, but provides a new out-of-the-box way to use Axes without having to override it if you don't use `ModelBackend`. 

The docs are updated to recommend using `AxesStandaloneBackend` because `ModelBackend` (or another backend that implements authentication) is still required. 

All tests pass. (The tests still use `AxesBackend`, not `AxesStandaloneBackend`. When they are set to use `AxesStandaloneBackend` instead, all tests pass except the `test_checks.py` ones specifically looking for `AxesBackend` to be used.)

### DIFF
--- a/axes/backends.py
+++ b/axes/backends.py
@@ -1,6 +1,6 @@
 from typing import Optional
 from django.conf import settings
-from django.contrib.auth.backends import ModelBackend
+from django.contrib.auth.backends import BaseBackend, ModelBackend
 from django.http import HttpRequest
 
 from axes.exceptions import (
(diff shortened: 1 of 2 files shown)


aleksihakli (approved):
LGTM

aleksihakli:
Released as part of 5.36.0

Issue #921, opened 2022-09-15
ImportError: cannot import name 'get_distribution' from 'pkg_resources'


jedie:
django_axes v5.39.0
```
  File ".../.venv/lib/python3.9/site-packages/axes/__init__.py", line 1, in <module>
    from pkg_resources import get_distribution
ImportError: cannot import name 'get_distribution' from 'pkg_resources' (unknown location)
```

PR #922, opened 2022-09-15
Bugfix #921 cannot import name 'get_distribution'

jedie:
Replace `pkg_resources` with `importlib` solution to fix https://github.com/jazzband/django-axes/issues/921


axes/__init__.py
```diff
@@ -1,4 +1,12 @@
-from pkg_resources import get_distribution
+def get_axes_version():
```
aleksihakli, line 1: Could we have this as plain top-level code without a function wrapping it?

The style checker disagrees with non-top-level imports and the QA test run won't hence pass.

aleksihakli (changes requested):

livenson:
Hi, it would be great if this could get in, we are trying to cleanup our codebase from setuptools and axes is a dependency. Can I help out somehow?

PR #933, opened 2022-10-25
add missing db alias to migration 0007

### LINKED ISSUE #932: Migration 0007 missing db alias
Migration [0007](https://github.com/jazzband/django-axes/blob/master/axes/migrations/0007_alter_accessattempt_unique_together.py) is missing the use of a database alias resulting in issues if the database option is used when applying.

As per [docs](https://docs.djangoproject.com/en/4.1/ref/migration-operations/#runpython) `db_alias = schema_editor.connection.alias` should be used.

liampauling:
closes #932


aleksihakli:
Thanks for the update!

PR #936, opened 2022-10-27
Add Python 3.11 to CI, tox, and trove classifiers

joshuadavidthomas:


.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', 'pypy-3.8']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11', 'pypy-3.8']
```
hramezani, line 16: I think we need to add exclude for Django 3.2 and 4.0

```
          - python-version: '3.11'
            django-version: '3.2'
          - python-version: '3.11'
            django-version: '4.0'

```

pyproject.toml
```diff
@@ -10,10 +10,10 @@ DJANGO_SETTINGS_MODULE = "tests.settings"
 legacy_tox_ini = """
 [tox]
 envlist =
-    py{37,38,39,310,py38}-dj32
-    py{38,39,310,py38}-dj40
-    py{38,39,310,py38}-dj41
-    py{38,39,310,py38}-djmain
+    py{37,38,39,310,311,py38}-dj32
```
hramezani, line 13: I think we don't need this change

pyproject.toml
```diff
@@ -10,10 +10,10 @@ DJANGO_SETTINGS_MODULE = "tests.settings"
...
+    py{38,39,310,311,py38}-dj40
```
hramezani, line 14: I think we don't need this change

aleksihakli:
Django - Python version support matrix is available at e.g. https://docs.djangoproject.com/en/4.1/faq/install/

hramezani (approved):

PR #950, opened 2022-11-24
Fix user agent

### LINKED ISSUE #930: AXES_USE_USER_AGENT not functioning correctly
This might be a duplicate of #677, but the user there did not provide enough detail for me to be sure.

I am using the following versions of Django and Axes:
```
django==3.2.16
django-ipware==4.0.2
django-axes==5.39.0
```

My Axes configuration is as follows:
```python
AXES_FAILURE_LIMIT = 5
AXES_COOLOFF_TIME = dt.timedelta(hours=1)
AXES_HANDLER = "axes.handlers.cache.AxesCacheHandler"
AXES_USERNAME_FORM_FIELD = "email"
AXES_USE_USER_AGENT = True
```

When I reach the 5 failure limit with a given `(ip, user_agent)` pair, I am locked out. If I change to a different user agent, I am still locked out. This is not expected, and the docs do not reflect this behaviour either. If I have `AXES_USE_USER_AGENT = True`, I would expect that using the same IP but a different user agent would reset the count.

It also appears that if, for example, I had 5 attempts with `(ip_1, user_agent_1)`, and then tried to log in with a **different** ip address (let's call it `ip_2`) but the same user agent string, I am *also* blocked out. It appears to lock out via user agent string, even if the IP is different!

Happy to provide more detail if necessary. I can also start digging into the code and/or provide some tests that prove these bugs exist, just haven't had the time yet...

PetrDlouhy:
Fix #930, add test

### DIFF
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -205,7 +205,7 @@ def get_client_parameters(username: str, ip_address: str, user_agent: str) -> li
 
         if settings.AXES_USE_USER_AGENT:
             # 4. The HTTP User-Agent can be used to track e.g. one browser
-            filter_query.append({"user_agent": user_agent})
+            filter_query[0]["user_agent"] = user_agent
 
     return filter_query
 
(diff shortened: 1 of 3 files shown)


aleksihakli:
Thanks, good catch @PetrDlouhy !

PR #952, opened 2022-12-02
Update prospector to 1.8.2

hramezani:


.prospector.yaml
```diff
@@ -5,3 +5,11 @@ ignore-paths:
...
   options:
     max-line-length: 142
+
+pylint:
+  disable:
+    - django-not-configured
+
+pyflakes:
+  disable:
+    - F401
```
hramezani, line 15: I had to add this here to ignore `F401`. not sure why `pyflakes` raises the below error even though we have `# noqa` on line 32

```
axes/apps.py
    Line: 32
      pyflakes: F401 / 'axes.checks' imported but unused (col 9)
```

aleksihakli:
LGTM, if it's possible to further configure the pyflakes check that'd be great, but disabling the check shouldn't produce that many problems either.

PR #967, opened 2022-12-29
Fix PyPy specifiers for tox and GH Actions

aleksihakli:


aleksihakli:
Yes, thanks for the cleanup.

It seems like PyPy specifiers were broken in some intermediate update but have since been fixed in the source package :)

PR #1033, opened 2023-04-14
Override log handler when using sensitive parameters.

### LINKED ISSUE #1010: `AXES_SENSITIVE_PARAMETERS` doesn't affect database handler logs
Hello!
I encountered an issue when attempting to utilize the `AXES_SENSITIVE_PARAMETERS` configuration in order to prevent usernames from being written into our log files, even though I added the setting like so:
```python
AXES_SENSITIVE_PARAMETERS = ["username"]
```
Whenever a login attempt fails a log is still being written:
image

I might've misunderstood how it works, but I did some debugging and found the following:
1. Inside `axes.handlers.database.py` some of the WARN logs are being written using the `client_str` variable
2. The `client_str` variable is produced from the `get_client_str` function in `axes.helpers.py`
3. In `get_client_str`, lines 298-302 should be replaced with a call to `get_query_str`, which does the same thing but also "cleanses" the dictionary, using the `AXES_SENSITIVE_PARAMETERS ` configuration value.

I'd love some feedback whether I misunderstood the usage of this configuration, or is this an actual issue,
Thanks! Shai.

stereodamage:
Closes #1010.

Not quite sure if it was needed in `AXES_VERBOSE` mode only, so made the changes afterwards.

### DIFF
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -306,7 +306,7 @@ def get_client_str(
         client_dict = {}
         for client in client_list:
             client_dict.update(client)
-
+    client_dict = cleanse_parameters(client_dict.copy())
     # Path info is always included as last component in the client string for traceability purposes
     if path_info and isinstance(path_info, (tuple, list)):
         path_info = path_info[0]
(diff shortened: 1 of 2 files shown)


aleksihakli:
Thank you for the fix 👍

Issue #1026, opened 2023-04-03
Compatiblity with django-auth-ldap?


joewesch:
Hello, I am inquiring to see if there has been any testing for compatibility with the [django-auth-ldap](https://django-auth-ldap.readthedocs.io/en/latest/index.html) library. I didn't see it in the [Integration documentation](https://django-axes.readthedocs.io/en/latest/6_integration.html). Has anyone tested this yet?

aleksihakli:
Hey!

In principle the package should be compatible if the `django-auth-ldap` authentication backend but it would be nice if you test it and confirm back.

In practice you can try to log in N times where N >= failure limit and check what happens.

PR #1035, opened 2023-04-18
Docs: Adds Django Auth LDAP as compatible

joewesch:
Closes: #1026 

I was able to verify that `django-auth-ldap` looks to be fully compatible with `django-axes`. I started with just adding `django-auth-ldap` as a control and to verify LDAP works. Afterward, I added `django-axes` and tested the IP lockout successfully - I got a lock out message after 3 invalid credentials. I continued to get the lock out message even refreshing and trying valid credentials again. I lastly tested that I was still able to log in via LDAP and local auth on another IP without issue.

### DIFF
--- a/docs/6_integration.rst
+++ b/docs/6_integration.rst
@@ -20,6 +20,7 @@ Django Allauth                                           |check|
 Django Simple Captcha                                    |check|
 Django OAuth Toolkit                                     |check|
 Django Reversion                                         |check|
+Django Auth LDAP                          |check|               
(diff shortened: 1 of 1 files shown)


aleksihakli:
That's great @joewesch! Thanks for testing the stack and updating the docs 👍

aleksihakli:
BTW if you have the time and opportunity it'd be great to add a section on `django-auth-ldap` to mention it works "out-of-the-box" if that's the case.

Issue #784, opened 2021-09-06
Increment the number of failed login attempts atomically in AxesCacheHandler


okapies:
`AxesCacheHandler` has a chance to increment `failures_since_start` badly where multiple worker processes receive many concurrent login attempts during the short period (e.g., DDoS). It requires to use of the atomic counter functionality inherent in the caching mechanism like [`INCR` in Redis](https://redis.io/commands/INCR), instead of implementing it ourselves:

https://github.com/jazzband/django-axes/blob/ac86d4b2139f632cee972e874984982a9628ab27/axes/handlers/cache.py#L100

Ref. #777.

aleksihakli:
This would be a very nice improvement if we could implement it in a way that is supported by at least the most common cache backends.

It is worth considering through that the users utilizing the cache handler due to number or volume of requests or DDoS resistance will likely use a specific cache such as Memcached or Redis, which are likely to support atomic incrementation of the cache values.

The Django cache framework does provide a `cache.incr` method:

https://docs.djangoproject.com/en/3.2/topics/cache/#django.core.caches.cache.incr

This is atomic on select backends which support the atomic incrementation and have implemented the correct logic in the cache layer.

aleksihakli:
Added `6.x` label for reference, can also be implemented in `5.x`.

aleksihakli:
This should be implemented by #1048

Issue #693, opened 2020-12-22
Version 6.0.0 release


aleksihakli:
The goal for the next major version 6 is to produce an improved version of Axes where the functionality and configuration is more consistent and documentation has more examples of configuring the project to suit individual project scenarios.

This issue will track larger design points and deprecating functionality as a whole and act as an entrypoint for people wanting to participate in the version 6 design, development, testing, and release process.

There is no official timeline or feature set agreed upon as of yet, the project is open for participation and discussion.

Larger features that will cause deprecation in settings and APIs and require the major semantic versioning bump to 6.x include.

- Improvement of HTTP status codes based on #570
- Improvement of sensitive data processing based on #663
- Redesign of lockout parameter resolution based on #690 and #692

aleksihakli:
Hey @hirotasoshu, do you happen to have thoughts on what to include in the initial `6.0.0` release?

I would kind of want to see #690 and #692 as part of the breaking release as they have been on the backlog for quite some time, any thoughts on those? I think specifying either a callable or a list of arguments to take into account would be a good approach?

aleksihakli:
All related issues and proposals are merged. The next would be to check the current docs and instructions, bake release notes, and write upgrade instructions in the docs for anyone updating from version 5.x to 6.x. Are there any other blockers for making the release after that?

hirotasoshu:
@aleksihakli I think we should also include #772 to new release. Since we already deprecate a lot of old stuff, deprecate `is_admin_site` will look harmonious in a major release

aleksihakli:
Agreed. I don't have time to update the implementations myself any time soon but would you or anyone else happen to have the chance of making a PR for that?

aleksihakli:
We should make release notes and upgrade instructions for people who wish to use Axes and keep the same behaviour as they had in the previous release. 

After that I think we could roll out version 6.0.0 as a latest stable release?

The RTD documentation and changelog could perhaps be a good place to include the instructions in.

aleksihakli:
Version 6.0 has been completed and released into PyPI along with patch 6.0.1 which includes fixes to CI and RTD configurations 🎉 

This issue can still be used to report any issues with the latest release you may encounter, or you can open separate issues and PRs as you see best.

PR #1036, opened 2023-04-24
Set 429 as default response code

### LINKED ISSUE #728: Would HTTP 429s be more appropriate?

hirotasoshu:
Resolves #728


aleksihakli:
I made a new `stable/5.x` release for maintaining backwards compatibility in release 5.x but this needs a new major 6.x release as it is a breaking change. I'm wondering if we should also include the changes proposed in https://github.com/jazzband/django-axes/issues/693 as well as make the `django-ipware` dependency optional as discussed in https://github.com/jazzband/django-axes/issues/1021 with the release `6.x`?

hirotasoshu:
@aleksihakli yep, I think these changes are suitable for the next release, I will create seperate issue for django-ipware. I think we should also include #784 because it can be a dealbreaker for highload services

aleksihakli:
Released as `6.0.0b1` and available as a prerelease in PyPI: https://pypi.org/project/django-axes/#history

aleksihakli:
Sure if #784 could be implemented I think it'd make a great candidate for the next release, especially a major breaking release.

Issue #1038, opened 2023-04-25
Make the `django_ipware` an optional dependency


hirotasoshu:
Discussed in #1021

aleksihakli:
Thanks for opening 👍 I made a prerelease for `6.0.0` marked as `6.0.0b1` to PyPI that we can test this in when implemented.

aleksihakli:
I sketched out one option for the optional deps pattern so that we have some basis for further discussion, would appreciate if you can check and highlight any issues or faults you detect in it. All improvement ideas are welcome :)

aleksihakli:
Well after some pondering the default idea in ipware, too, seems to be to just use `request.META["REMOTE_ADDR"]` if nothing better is configured and available, and for ease-of-use I'd assume that defaulting to that if we make `django-ipware` optional would be a good option? I changed the PR draft to use that approach, do check if it seems OK.

aleksihakli:
This has been implemented and published in the prerelease version 6.0.0b2 to enable feature testing.

PR #1044, opened 2023-04-26
Make ipware an optional dependency

aleksihakli:
Relates to #1038

### DIFF
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -5,7 +5,6 @@ from string import Template
 from typing import Callable, Optional, Type, Union
 from urllib.parse import urlencode
 
-import ipware.ip
 from django.core.cache import BaseCache, caches
 from django.http import HttpRequest, HttpResponse, JsonResponse, QueryDict
 from django.shortcuts import redirect, render
@@ -15,6 +14,13 @@ from axes.conf import settings
 from axes.models import AccessBase
 
 log = getLogger(__name__)
+
+try:
+    import ipware.ip
+
+    IPWARE_INSTALLED = True
+except ImportError:
+    IPWARE_INSTALLED = False
 
 
 def get_cache() -> BaseCache:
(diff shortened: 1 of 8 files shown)


axes/helpers.py
```diff
@@ -148,6 +148,24 @@ def get_client_username(
     return request_data.get(settings.AXES_USERNAME_FORM_FIELD, None)
 
 
+def get_client_ip_address_default(request: HttpRequest) -> str:
+    try:
+        import ipware.ip
+    except ImportError as e:
+        raise AxesBackendImplementationRequired(
+            "Axes requires either configuring settings.AXES_CLIENT_IP_CALLABLE django-ipware package to be installed with e.g. pip install django-axes[ipware]."
+        ) from e
```
hirotasoshu, line 157: This try/except block will cause python to try to import `ipware` during each function call. Instead, you can do smth like that in head of the module:
```python
try:
    import ipware.ip
    IPWARE_INSTALLED = True
except ImportError:
     IPWARE_INSTALLED = False
```
and in function body:
```python
if not IPWARE_INSTALLED:
    raise ...

File: axes/helpers.py [hunk 2]
```diff
@@ -148,6 +148,24 @@ def get_client_username(
...
+
+    client_ip_address, _ = ipware.ip.get_client_ip(
+        request,
+        proxy_order=settings.AXES_PROXY_ORDER,
+        proxy_count=settings.AXES_PROXY_COUNT,
+        proxy_trusted_ips=settings.AXES_PROXY_TRUSTED_IPS,
+        request_header_order=settings.AXES_META_PRECEDENCE_ORDER,
```
hirotasoshu, line 164: As you said in #1038, I think we should implement some logic to support old values and add warning that this values will be deprecated in 6.1 or 6.2 to ensure a smoother transition to new release

hirotasoshu (changes requested):

axes/helpers.py
aleksihakli: Yeah, let's do that 👍

axes/helpers.py
aleksihakli: Deprecation warnings added

setup.py
```diff
@@ -36,7 +36,10 @@
     use_scm_version=True,
     setup_requires=["setuptools_scm"],
     python_requires=">=3.7",
-    install_requires=["django>=3.2", "django-ipware>=3", "setuptools"],
+    install_requires=["django>=3.2", "setuptools"],
+    extras_require={
+        "ipware": "django-ipware>=3",
```
aleksihakli, line 41: Wondering if this should be `ipware`,  `django-ipware` or both in the `extras` section.

hirotasoshu: It seems to me that `ipware` is enough, this is more brief, especially in the context of django it seems clear to me that `django-ipware` is meant

hirotasoshu (approved):
LGTM, great work! I left one comment, if you have a chance to add a new test that would be great

setup.py
aleksihakli: Yeah let's keep it just `ipware`. It's explicit enough and it'll be easy to add more specifiers later if needed 👍

axes/helpers.py
```diff
@@ -173,15 +179,22 @@ def get_client_ip_address(request: HttpRequest) -> str:
...
 
-    return client_ip_address
+        client_ip_address, _ = ipware.ip.get_client_ip(
+            request,
+            proxy_order=settings.AXES_IPWARE_PROXY_ORDER,
+            proxy_count=settings.AXES_IPWARE_PROXY_COUNT,
+            proxy_trusted_ips=settings.AXES_IPWARE_PROXY_TRUSTED_IPS,
+            request_header_order=settings.AXES_IPWARE_META_PRECEDENCE_ORDER,
+        )
+        return client_ip_address
```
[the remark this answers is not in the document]
aleksihakli: That seems like a good idea for the test stack, I'll update the stack to include the tests for running with/without `ipware`

PR #1048, opened 2023-04-28
Use cache.incr for setting the cached authentication failure numbers

aleksihakli:
Relates to #784

Notes on implementation:

- All counters are not in the same value; if we track login attempts by multiple attributes such as username and/or IP address one counter can be higher, one can be lower, if someone is failing to authenticate from multiple IPs with different usernames or with multiple usernames from the same IP. Simplifying solution to only support exclusive username XOR IP address based tracking (instead of include OR) would enable just tracking the user via single attribute or a composite attribute.
- The current cache key(s) calculation approach is not exactly pretty, but it is compatible with the database implementation?

### DIFF
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -259,7 +259,7 @@ def make_cache_key_list(filter_kwargs_list):
     return cache_keys
 
 
-def get_client_cache_key(
+def get_client_cache_keys(
     request_or_attempt: Union[HttpRequest, AccessBase],
     credentials: Optional[dict] = None,
 ) -> str:
--- a/axes/helpers.py
+++ b/axes/helpers.py
@@ -2,7 +2,7 @@ from datetime import timedelta
 from hashlib import sha256
 from logging import getLogger
 from string import Template
-from typing import Callable, Optional, Type, Union
+from typing import Callable, Optional, Type, Union, List
 from urllib.parse import urlencode
 
 from django.core.cache import BaseCache, caches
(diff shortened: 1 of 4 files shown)


hirotasoshu:
> * All counters are not in the same value; if we track users by multiple attributes such as username and/or IP address one tracker can be higher, one can be lower, if someone is attempting to authenticate from multiple IPs with different usernames or with multiple usernames from the same IP. Simplifying solution to only support exclusive username XOR IP address based tracking (instead of include OR) would enable just tracking the user via single attribute or a composite attribute.

As I can see in current implementation, if we use username AND ip address, only one cache key is created so we have only one tracker already.
For situation, when we use username OR ip address, I think current implementation is correct, because with that implementation we can lockout users that tries to login with different ips and same username AND users that tries to login with different usernames and same ip AND users that tries to login with same username and same ip. If we use only one attribute or composite attribute we can't cover all that cases. The only problem is that this approach can lead to the creation of a large number of keys, which in turn can lead to high RAM consumption in the case of using IN-MEMORY databases. But it seems to me that in this case you just need to configure it correctly.

axes/helpers.py
```diff
@@ -259,7 +259,7 @@ def make_cache_key_list(filter_kwargs_list):
     return cache_keys
 
 
-def get_client_cache_key(
+def get_client_cache_keys(
     request_or_attempt: Union[HttpRequest, AccessBase],
     credentials: Optional[dict] = None,
 ) -> str:
```
hirotasoshu, line 265: Since we returning list of keys, type hint should be list[str]

hirotasoshu (approved):
LGTM!

axes/helpers.py
aleksihakli: Yeah good catch. I was looking at the docstrings, naming and type hints and they could use a check-up to verify correctness. I've become a bit too accustomated to the code base and am not catching them with a fresh eye anymore.

aleksihakli:
Released in version 6.0.0b3

Issue #690, opened 2020-12-17
No flexibility for get_client_parameters (AXES_LOCK_OUT_BY_COMBINATION_USER_AGENT_AND_IP, AXES_LOCK_OUT_BY_COMBINATION_USER_AGENT_AND_USER and AXES_LOCK_OUT_BY_USER_AGENT_ONLY options needed)


instantrad:
We realised with below config; Every user on our website gets locked out. 
AXES_FAILURE_LIMIT = 2
AXES_LOCK_OUT_AT_FAILURE = True
ONLY_USER_FAILURES = False
AXES_USE_USER_AGENT = True
AXES_COOLOFF_TIME = timedelta(minutes=15)

The issue is that the **filter_query** filter out the ip or user_agent. This means that all users with **same IP** but different **user agents** get locked out.

**PROPOSAL:**

```def get_client_parameters(username: str, ip_address: str, user_agent: str) -> list:
    if settings.AXES_ONLY_USER_FAILURES:
        filter_query = [{"username": username}]
    else:
        if settings.AXES_LOCK_OUT_BY_USER_OR_IP:
            filter_query = [{"username": username}, {"ip_address": ip_address}]
        elif settings.AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP:
            filter_query = [{"username": username, "ip_address": ip_address}]
        elif settings.AXES_LOCK_OUT_BY_COMBINATION_USER_AGENT_AND_IP:
            filter_query = [{"user_agent": user_agent, "ip_address": ip_address}]
        elif settings.AXES_LOCK_OUT_BY_COMBINATION_USER_AGENT_AND_USER:
            filter_query = [{"user_agent": user_agent, "username": username}]
        elif settings.AXES_LOCK_OUT_BY_USER_AGENT_ONLY:
            filter_query = [{"user_agent": user_agent}]
        else:
            filter_query = [{"ip_address": ip_address}]
    return filter_query```

aleksihakli:
I have a wish of cleaning up the flags and just supplying users the option of configuring the filter parameters that should be in the query themselves by keywords for example. 

Something like:

    # settings.py
    AXES_LOCKOUT = ["ip_address"]

Then just check that configuration on startup and maybe even allow users to supply their own custom method of resolving the lockout parameters from the request.

Would it make sense to refactor the whole implementation? It's messy and has layers of legacy to be honest.

instantrad:
[...]

PR #1050, opened 2023-05-04
Refactor client lockout parameter resolution

### LINKED ISSUE #692: Refactor client lockout parameter resolution
Opening public discussion on feature request for redesigning and refactoring the lockout parameter resolution functionality as per discussion in jazzband/django-axes#690

Proposal: use a new flag such as `AXES_LOCK_OUT_PARAMETERS` which accepts an iterable of keys or a callable that resolves an iterable. Use a sane default value with behaviour similar to that of current defaults. Make a new major release version 6 release that deprecates the old lock out parametrization. Also go through other flags that could be cleaned up?

Example:

```
# settings.py
# this could also be a callable that takes request and credentials as parameters?
AXES_LOCKOUT_PARAMETERS = ["ip_address"]
```

Deprecate or clean up the following flags ([reference](https://django-axes.readthedocs.io/en/latest/4_configuration.html#configuring-project-settings)):

- AXES_USE_USER_AGENT
- AXES_LOCK_OUT_AT_FAILURE
- AXES_LOCK_OUT_BY_USER_OR_IP
- AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP

hirotasoshu:
Resolves #692 and #690


axes/utils.py
```diff
@@ -37,23 +37,38 @@ def reset_request(request: HttpRequest) -> int:
...
+            ip_required = True
+
+    ip_or_username = not ip_and_username and ip_required and username_required
+    if not ip_required:
         ip = None
-    elif not (
-        settings.AXES_LOCK_OUT_BY_USER_OR_IP
-        or settings.AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP
-    ):
+    if not username_required:
```
hirotasoshu, line 66: That's kindy ugly, I think we should change reset API

axes/apps.py
```diff
@@ -33,14 +33,22 @@ def initialize(cls):
...
-                mode = "blocking by IP only"
+                mode = "blocking by " + " or ".join(
+                    [
+                        param
+                        if isinstance(param, str)
+                        else "combination of " + " and ".join(param)
+                        for param in settings.AXES_LOCKOUT_PARAMETERS
+                    ]
+                )
+
```
hirotasoshu, line 51: This produces almost the same log messages as we had before (the only difference that it IP word changed to ip_address and messages does not have word only if we have only one lockout parameter)

axes/handlers/cache.py
```diff
@@ -78,9 +79,10 @@ def user_login_failed(self, sender, credentials: dict, request=None, **kwargs):
             return
 
         username = get_client_username(request, credentials)
-        if settings.AXES_ONLY_USER_FAILURES and username is None:
+        lockout_parameters = list(get_lockout_parameters(request, credentials))
```
hirotasoshu, line 82: Since AXES_LOCKOUT_PARAMETERS could be any iterable, for simplicity, I convert it to list

axes/helpers.py
```diff
@@ -226,26 +255,35 @@ def get_client_parameters(username: str, ip_address: str, user_agent: str) -> li
...
+                    for combined_parameter in parameter
+                }
+            filter_kwargs.append(filter_kwarg)
+
+        except KeyError as e:
+            error_msg = (
+                f"{e} lockout parameter is not allowed. "
+                f"Allowed parameters: {', '.join(parameters_dict.keys())}"
+            )
+            raise ValueError(error_msg) from e
```
hirotasoshu, line 284: Should we raise error like this or just skip incorrect parameters?
Or maybe we can check parameters for correctness in `get_lockout_parameters`?

tests/test_attempts.py
```diff
@@ -82,74 +82,74 @@ def test_reset_ip_username(self):
...
         self.assertEqual(AccessAttempt.objects.count(), 3)
 
-    @override_settings(AXES_LOCK_OUT_BY_USER_OR_AND=True)
+    @override_settings(AXES_LOCKOUT_PARAMETERS=[("username", "ip_address")])
     def test_reset_ip_username_user_and_ip(self):
         self.request.GET["username"] = self.USERNAME_1
         self.request.META["REMOTE_ADDR"] = self.IP_1
         reset_request(self.request)
-        self.assertEqual(AccessAttempt.objects.count(), 3)
+        self.assertEqual(AccessAttempt.objects.count(), 4)
```
hirotasoshu, line 155: This test was incorrect

hirotasoshu:
Tests broke in old python versions, because `Iterable` before python3.8 was in collections module, not collections.abc

hirotasoshu:
@aleksihakli I left a couple comments for this implementation. If everything is seems ok to you, I will add new tests (since now lockout parameters and we need additional tests to ensure everything is working fine) and update documentation

aleksihakli:
Thanks for the PR! The approach looks good and backwards compatible which is nice. I'll try and wrap up a review soon.

aleksihakli:
Looks good to me.

I'd maybe adjust a few small things either before or after merging, it's easy to make separete update commits or PRs on top of this after upstreaming too:

- change the types for the: use nested lists of instead of tuples inside lists, just a personal preference for the sake of simplicity
- add a `log.exception` message on invalid lockout keys / values for sake of debuggability
- add a test for triggering the `KeyError` / `ValueError` for sake of verification (maybe test the log trace as well in the same place)
- add a deprecation check for anyone that might use the new and old flags in a mixed fashion by accident, and
- add a few more docstring examples for the new lockout param introductions and available variants

but these are just small nuances and preferences.

I think this PR implements the setting and API updates in a more flexible and more understandable fashion than the current stack does as it currently is. It's a good pragmatic approach for modernizing the implementation 👍

Thanks for the effort. Let's get this merged and update instructions written so we can roll out 6.0 soon 💪

aleksihakli:
Current `prospector` errors to-fix:

https://github.com/jazzband/django-axes/actions/runs/4881413414/jobs/8710287094?pr=1050#step:7:91

Current typing errors to-fix:

https://github.com/jazzband/django-axes/actions/runs/4881413414/jobs/8710285656?pr=1050#step:7:154

hirotasoshu:
@aleksihakli I adjusted types for ``AXES_LOCKOUT_PARAMETERS``: it's now list of strings or nested lists of strings, added log.exception for cases when incorrect lockout parameters passed and also added new tests

aleksihakli:
Thank you for the effort, this is a great update to get into the 6.x release to refresh the API and settings and it even has backwards compatibility which is awesome👍I'll go ahead and merge this and bake a new beta release tag, let's see if we have anything to add after that before rolling out 6.0.0 🎉

PR #1054, opened 2023-05-13
Fix inaccuracies in russian locale

### LINKED ISSUE #1051: inaccuracies in russian locale
As a native russian speaker, I found some inaccuracies in Russian translation:
1.
https://github.com/jazzband/django-axes/blob/b985981d119793f4b53d66fb9a5db201e3127915/axes/locale/ru/LC_MESSAGES/django.po#L56-L57
Запрос HTTP translates literally as an HTTP request and it is not correct since HTTP Accept is HTTP header
2. 
https://github.com/jazzband/django-axes/blob/b985981d119793f4b53d66fb9a5db201e3127915/axes/locale/ru/LC_MESSAGES/django.po#L44-L45
Браузер пользователя translates literally as an user browser and it is not always correct (curl for example is not browser, but by default requests made from curl have user agent curl/x.x.x)

I think we can't translate it properly to russian language without losing meaning and should keep it as it is (like in polish locale)

3. 
https://github.com/jazzband/django-axes/blob/b985981d119793f4b53d66fb9a5db201e3127915/axes/locale/ru/LC_MESSAGES/django.po#L36-L41
I don't think that word "разблокирования" exists in Russian language. It would be more correct to restructure this sentence in this way:
Свяжитесь с администратором, чтобы разблокировать учетную запись

hirotasoshu:
Resolves #1051


aleksihakli:
Thanks for the updates!

PR #1057, opened 2023-05-16
Version 6

### LINKED ISSUE #693: Version 6.0.0 release

aleksihakli:
Proposal for next major breaking release version 6.0.0 with upgrade instructions from version 5.x.

Also dropping Python 3.7 support because it is EOL in 1 month and the setup is simpler without it going forward.

Closes #693

### DIFF
--- a/docs/2_installation.rst
+++ b/docs/2_installation.rst
@@ -34,10 +34,10 @@ After installing the package, the project settings need to be configured.
         'django.contrib.auth.backends.ModelBackend',
     ]
 
-    For backwards compatibility, ``AxesBackend`` can be used in place of ``AxesStandaloneBackend``. 
+    For backwards compatibility, ``AxesBackend`` can be used in place of ``AxesStandaloneBackend``.
     The only difference is that ``AxesBackend`` also provides the permissions-checking functionality
     of Django's ``ModelBackend`` behind the scenes. We recommend using ``AxesStandaloneBackend``
-    if you have any custom logic to override Django's standard permissions checks. 
+    if you have any custom logic to override Django's standard permissions checks.
 
 **3.** Add ``axes.middleware.AxesMiddleware`` to your list of ``MIDDLEWARE``::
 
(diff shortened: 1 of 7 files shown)


docs/2_installation.rst
```diff
@@ -75,6 +75,70 @@ Many people have different configurations for their development and production e
...
+- ``settings.AXES_LOCK_OUT_BY_USER_OR_IP``, and
+- ``settings.AXES_USE_USER_AGENT``.
+
+To keep old behaviour with the new flag, configure the following:
+
+- If you did not use any flags, use ``settings.AXES_LOCKOUT_PARAMETERS = ["ip_address"]``,
+- If you used ``settings.AXES_ONLY_USER_FAILURES``, use ``settings.AXES_LOCKOUT_PARAMETERS = ["username"]``,
+- If you used ``settings.AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP``, use ``settings.AXES_LOCKOUT_PARAMETERS = [["username", "ip_address"]]``,
+- If you used ``settings.AXES_LOCK_OUT_BY_USER_OR_IP``, use ``settings.AXES_LOCKOUT_PARAMETERS = ["username", "ip_address"]``, and
+- If you used ``settings.AXES_USE_USER_AGENT``, add ``"user_agent"`` to your list(s) of lockout parameters.
```
hirotasoshu, line 121: As for use_user_agent, it should be clarified:
1. If other flags were not used, then you need to configure like this
```python
 settings.AXES_LOCKOUT_PARAMETERS = [["ip_address", "user_agent"]]
 ```
 2. If there was a flag ONLY_USER_FAILURES, then 
 
```python
 settings.AXES_LOCKOUT_PARAMETERS = [["username", "user_agent"]]
 ```
 3. If there was a flag AXES_LOCKOUT_BY_USER_OR_IP:
 
```python
 settings.AXES_LOCKOUT_PARAMETERS = [["ip_address", "user_agent"], "username"]
 ```
 4. If there was a flag AXES_LOCKOUT_BY_COMBINATION_USER_AND_IP, than
  
```python
 settings.AXES_LOCKOUT_PARAMETERS = [["ip_address", "user_agent", "username"]]
 ```
 This needs to be clarified, because how it worked before is not at all obvious

aleksihakli: Default for `AXES_USE_USER_AGENT` was previously `False` and the defaults as they are defined today are described in https://github.com/jazzband/django-axes/blob/d463787e343f80234e392435920a51a47bfe19a9/axes/conf.py#L17

`user_agent` is only used if `AXES_USE_USER_AGENT` is `True`, and it is not included by default in the settings?

aleksihakli: Updated, does this look better now?

Issue #1058, opened 2023-05-22
Callables in settings are not verified at start-up if they are strings


iafisher:
If you accidentally set, e.g., `AXES_LOCKOUT_CALLABLE` to an invalid import path, the server will still boot up fine and you won't find out about it until Axes tries to invoke the callable [here](https://github.com/jazzband/django-axes/blob/fcdeb4ae3f11724b99da54dfa61df90abcf54f7b/axes/helpers.py#L453). In my case, `AXES_LOCKOUT_CALLABLE` was misconfigured for months before I saw the error in the logs.

It would be ideal if Axes did some kind of verification of the callable settings at start-up. I can think of two ways:

- Verify settings in `conf.py`
- Verify settings in a custom [system check](https://docs.djangoproject.com/en/4.1/topics/checks/)

If this seems like a good idea to you I'd be willing to work on it. Not sure if crashing the server on misconfiguration counts as a backwards-incompatible change or not; if so could print a warning instead.

I am using version 5.39.0.

aleksihakli:
Thanks for reporting!

This would also be a good opportunity to refactor the callable resolution to use one common function to return one verified callable object from a string path or an object itself. At the moment there is duplicate code in many places and it would be nice to clean that up.

@iafisher and @hirotasoshu would you happen to be interested about improving this and possibly baking a new release as you are familiar with the code base?

hirotasoshu:
@aleksihakli should it be new checks in `axes/checks.py`?

iafisher:
@hirotasoshu I'd be happy to write the system check if you want to work on the common function for resolving callables. I can get it done today or tomorrow.

hirotasoshu:
> @hirotasoshu I'd be happy to write the system check if you want to work on the common function for resolving callables. I can get it done today or tomorrow.

Ok, I created #1068 for this

aleksihakli:
Thank you for the work on the PR 👍 Upstreamed, I'll bake a new version to PyPI as well.

aleksihakli:
Released in version 6.0.2.

PR #1069, opened 2023-05-30
Add check for callable settings

iafisher:
Fixed: #1058


axes/checks.py
```diff
@@ -153,3 +156,47 @@ def axes_deprecation_check(app_configs, **kwargs):  # pylint: disable=unused-arg
...
+                    hint=Hints.CALLABLE_INVALID,
+                    id=Codes.CALLABLE_INVALID,
+                )
+            )
+
+    return warnings
+
+
+def is_valid_callable(value) -> bool:
+    if value is None:
```
hirotasoshu, line 188: As I mentioned in previous comment, we have settings that can take both normal values and be callable, so that check is not always valid

hirotasoshu (commented):
I think it can be safely upstreamed after adding checks for `AXES_COOLOFF_TIME` and `AXES_LOCKOUT_PARAMETERS`

iafisher:
Is there anything more you need me to do before this PR can be merged?

aleksihakli:
Looks good, thank you for the improvement!

PR #1080, opened 2023-06-20
Remove unused backend methods

### LINKED ISSUE #998: `force_login` not working in tests
Hey,

I've tried following the documentation and setting `AXES_ENABLED = False` in the test settings, but that did not work. I also tried removing 'axes' from INSTALLED_APPS in my test settings, but that also did not work?

I was able to get the test to work by using `client.login` and passing in a mock request, but it would be nicer if I could get it to work by just setting `AXES_ENABLED`.

My test:
``` python
@override_settings(AXES_ENABLED=False)
def test_gives_non_cas_admin_403_on_cas_admin_url(client, django_user_model):
    user = django_user_model.objects.create()

    client.force_login(user)
    response = client.get(reverse('create_user'))

    assert response.status_code == 403
```
The test is failing because it redirect back to the login view since the `force_login` is not working

314eter:
# What does this PR do?

`AxesStandaloneBackend` was inheriting a default implementation of `get_user()` from `BaseBackend`. Django's `force_login()` uses the first authentication backend that has a `get_user()` method for authorization (since django/django#7634), which fails because `AxesStandaloneBackend` will always be first but can't authorize.

The existing tests didn't notice this, because they never try to do a request after `force_login()`.

By removing the `BaseBackend` parent class, `AxesStandaloneBackend` doesn't have a `get_user()` method anymore, and will be skipped by `force_login()`. Since `AxesStandaloneBackend` never authorizes a user, nothing will ever try to call the inexisting `get_user()`.

This is the same solution [django-rules](https://github.com/dfunckt/django-rules) uses.

Fixes #998.

## Before submitting

### DIFF
--- a/tests/test_login.py
+++ b/tests/test_login.py
@@ -28,7 +28,7 @@ class DjangoLoginTestCase(TestCase):
         self.username = "john.doe"
         self.password = "hunter2"
 
-        self.user = get_user_model().objects.create(username=self.username)
+        self.user = get_user_model().objects.create(username=self.username, is_staff=True)
         self.user.set_password(self.password)
         self.user.save()
         self.user.backend = "django.contrib.auth.backends.ModelBackend"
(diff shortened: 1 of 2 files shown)


tests/test_login.py
```diff
@@ -28,7 +28,7 @@ def setUp(self):
         self.username = "john.doe"
         self.password = "hunter2"
 
-        self.user = get_user_model().objects.create(username=self.username)
+        self.user = get_user_model().objects.create(username=self.username, is_staff=True)
```
aleksihakli, line 31: Why has this been changed?

aleksihakli (approved):
LGTM. Why has the test user been changed to a staff user, though?

tests/test_login.py
314eter: Only staff users have access to `"admin:index"`. For normal users there is no difference between a failed and successful login.

aleksihakli: Sounds good, thank you.

aleksihakli:
Released as 6.0.4.

PR #1094, opened 2023-07-26
#438: Added docs about GDPR best practice

### LINKED ISSUE #438: GDPR discussion
Hi guys,

just talked with my colleage about the issue that with the new EU GDPR law you are basically not allowed to save IP addresses anywhere.

We came up with the idea to make a settings variable. If this is enabled, IP addresses get saved as a hash.

Any thoughts about that solution?

Best from Europe
Ronny

GitRon:
# What does this PR do?

Added documentation about GDPR best practices.

Fixes #438 

## Before submitting

cc @mastacheata @aleksihakli


GitRon:
@aleksihakli Do you know why the pipeline is failing? I didn't change any code...

aleksihakli:
I'll leave this open for a while for review & discussion opportunities :)

hirotasoshu:
@aleksihakli  @GitRon An alternative approach would be to hash the ip address, but this would require changing the type of the ip_address field from GenericIpAddressField to CharField. This will allow you to block requests based on both username and ip address, while not saving the real ip, however, under heavy loads it can lead to a large cpu load

docs/3_usage.rst
```diff
@@ -107,3 +107,24 @@ In your code, you can use the ``axes.utils.reset`` function.
...
+
+Data privacy and GDPR
+^^^^^^^^^^^^^^^^^^^^^
+
+Most European countries have quite strict laws regarding data protection and privacy. It's highly recommended and good 
+practice to treat your sensitive user data with care. The general rule here is that you shouldn't store what you don't need.
+
+When dealing with brute-force protection, the IP address and the username (often the email address) are most crucial.
+Given that you can perfectly use `django-axes` without locking the user out by IP but by username, it does make sense to
+avoid storing the IP address at all. You can not lose what you don't have.
```
aleksihakli, line 121: Afaik IP address is not sensitive user data or personal information per-se, if it's not combined with other identifiable data that points to a single user. And, on the other hand, we are storing login attempts which can be made by anyone on the world wide web. What data needs to be stored in what scenarios is a pretty interesting question that would be worth discussing a bit more about, but I'm not an expert on this field, so maybe it's easiest to let people open an issue ticket or a PR if they seem room for improvement on the instructions of use or any other points :)

aleksihakli (approved):

hirotasoshu:
> Maybe worth a settings flag to enable if you don't care about the CPU load?
I think we can just mention in docs that hashing IP address in records can be done via custom database handler)
Also we can mention that using cache handler is safe because it's using hash under the hood

docs/3_usage.rst
GitRon: Great idea! And thx for your effort! 👌

PR #1095, opened 2023-07-27
#1093: "AXES_SENSITIVE_PARAMETERS" now have  sensible defaults to fol…

### LINKED ISSUE #1093: FEATURE REQUEST: Logging without sensitive data by default (privacy by design)
Hi @aleksihakli 

we found out that AXES logs IP and username to the configured Django log on a failed attempt. The IP address is critical by definition but the username is very often the email address of the user - so it's even more critical.

I would love to see a solution where we avoid logging any sensitive information - unless you explicitly want this and enable it.

So my suggestion:
* `AXES_VERBOSE` is set to `False` by default
* The `get_client_str` method does not print the username when verbose mode is off but the User ID. This will help you as much as the username to find the specific user but you avoid spilling the data everywhere (like some fancy server or cloud logs)

What do you think? Would you be open for a PR?

Best  
Ronny

GitRon:
…low "privacy-by-design"

# What does this PR do?

The new `AXES_SENSITIVE_PARAMETERS` enables you to mask sensitive data in your logs. To follow privacy-by-design (zero config to be as compliant as possible) I added the two most critical fields. The database storage is not changed by this behaviour, you only avoid spilling potentially sensitive user-related data to your logs (which might end up somewhere).

If you need the logs for any reason, just set the variable in your project to an empty list and you are good to go.

Why I like this solution:
* Most people don't worry about this topic, they are presented with a secure solution
* If you need something else, you can VERY easily customise it the way you want it to be

Drawbacks:
* People who rely on the logs to track their access attempts have to configure the settings variable. But this change will be announced in the changelog and the amount of work is < 5min. In a nutshell, I believe that the benefits outweigh the drawbacks by far.

Fixes #1093 

## Before submitting


GitRon:
Hi @aleksihakli 

after going back and forth on this topic, I do beleive to have found a neat solution which serves in everybodys best interest.

Thx for reviewing!
Ronny

cc @mastacheata

aleksihakli:
I'll let users review this changeset so we can have more than one opinion on backwards compatibility and other notes. @camilonova @hramezani @hirotasoshu do you have any thoughts on this btw?

hirotasoshu:
@aleksihakli I don't see any problems that could arise with this update, because it will really take around one minute to return everything as it was before

GitRon:
@hirotasoshu Sounds great!

GitRon:
@aleksihakli Thx for the quick release! But the docs on RTD still seem to have the old version? Or am I missing something?

aleksihakli:
The documentation build seems to be failing:

https://readthedocs.org/projects/django-axes/builds/

Would anyone be interested in fixing the build?

aleksihakli:
@camilonova would it be possible to get access to RTD management page so I could update the build configuration?

https://readthedocs.org/projects/django-axes/

camilonova:
I don't have access. I guess @jezdez have it.

aleksihakli:
Roger that 👍 The current RTD documentation build fails because the `black` version is too new and doesn't support the build Python version. Updating the Python to e.g. 3.11 or a similarly new version would probably help.

GitRon:
@aleksihakli How and where would we fix that?

GitRon:
@aleksihakli Can I do something to move this forward? 🙂

aleksihakli:
You can view the build statuses at:

https://readthedocs.org/projects/django-axes/builds/

To fix the build the easiest option is to probably downgrade the `black` package to a version that is compatible with the Python version that RTD uses to build the docs.

aleksihakli:
I added a baseline `.readthedocs.yaml` configuration with basic Python 3.11 build configuration so the RTD docs should now be fixed and be configurable via the repository. Feel free to open any necessary PRs to improve the docs builds if you notice anything missing.

aleksihakli:
PDF build still seems to have issues, see https://github.com/jazzband/django-axes/issues/1102

If anyone has the opportunity to fix that it would be appreciated.

PR #1105, opened 2023-08-17
#1093: Setting proper headline (h2) for GDPR section

GitRon:
Hi @aleksihakli 

just saw at RTD that I used the wrong headline (h3 instead of h2). Fixed this issue.

Best  
Ronny


aleksihakli:
Thank you 👍

PR #1109, opened 2023-08-31
Fix `TransactionManagementError`

### LINKED ISSUE #1084: TransactionManagementError when writing database is not "default"
When starting atomic transactions in `AxesDatabaseHandler`, the database is not specified. This can cause a `TransactionManagementError` if the writing database is not the default one after `select_for_update()` is called.

The solution would be to call `db.router.db_for_write()` to get the database name:

```
with transaction.atomic(using=db.router.db_for_write(AccessAttempt)):
    (
        attempt,
        created,
    ) = AccessAttempt.objects.select_for_update().get_or_create(
...
```

```[tasklist]
### Tasks
```

hirotasoshu:
# What does this PR do?

Fixes #1084

## Before submitting

### DIFF
--- a/axes/handlers/database.py
+++ b/axes/handlers/database.py
@@ -1,8 +1,8 @@
 from logging import getLogger
 from typing import Optional
 
-from django.db import transaction
-from django.db.models import F, Sum, Value, Q
+from django.db import router, transaction
+from django.db.models import F, Q, Sum, Value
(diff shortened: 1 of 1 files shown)


aleksihakli (approved):
Thank you for the fix 👍

aleksihakli:
Released as 6.1.1

PR #1113, opened 2023-09-06
Update `is_valid_callable` check

sobolevn:
Changes:
1. We now check that type of an imported object (you can specify a wrong name)
2. By default we return `False`, because only `str` / `callable` / `None` are valid in this context, nothing else

## Before submitting

### DIFF
--- a/axes/checks.py
+++ b/axes/checks.py
@@ -197,8 +197,8 @@ def is_valid_callable(value) -> bool:
 
     if isinstance(value, str):
         try:
-            import_string(value)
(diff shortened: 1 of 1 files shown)


aleksihakli:
Hey @sobolevn. The PR seems to break the tests, would you happen to have the opportunity of checking and adjusting those as well?

sobolevn:
@aleksihakli sorry, I moved on from this :(
Please, feel free to close / update / fix.

PR #1115, opened 2023-09-07
Avoid running data migration on incorrect databases

christianbundy:
## Problem

When running Django with multiple databases, `./manage.py migrate --database other` fails because none of the Axes migrations run, but the data migration _does_ run, and fails because the tables don't exist.

## Solution

Early exit if the migrations are being run on a database where Axes doesn't exist.


aleksihakli:
Good catch, thank you for the PR 👍

PR #1121, opened 2023-09-20
Andreas dickow patch 1

AndreasDickow:
# What does this PR do?

the pull request fixes a problem with master-master replication database setups, as often found in large scale and fail safe systems.
A race condition by database function induced timestamp generation is circumvented by avoiding django auto_add and auto_add_now options

Fixes # (issue)

## Before submitting


aleksihakli:
I'm closing this as there has been no discussion on the implementation alternatives. I'm not opposed to adding this feature but it would be nice to discuss implementing this functionality without overwriting model methods and adding data layer logic.

Issue #1123, opened 2023-09-28
BUG: Middleware Support for Async Required to Adhere to Django’s Asynchronous Standards


krystofbe:
**Describe the bug** 

When using django-axes version 6.1.1, I encountered a warning related to async adaptation. According to Django's documentation on async safety (link: [Async Safety](https://docs.djangoproject.com/en/4.2/topics/async/#async-safety)), middleware can be built to support both sync and async contexts. Unfortunately, when using django-axes, I received the following warning: "Asynchronous handler adapted for middleware axes.middleware.AxesMiddleware." This indicates that the middleware is not natively supporting async contexts, which is not in line with the best practices outlined in Django's documentation.

**To Reproduce** 

1.  Install django-axes version 6.1.1 on a project running with Django supporting async features.
2.  Enable debug logging for the django.request logger.
3.  Make a request to the application.
4.  Observe the warning message indicating that the asynchronous handler was adapted for middleware axes.middleware.AxesMiddleware.

**Expected behavior** 

The expected behavior is that django-axes middleware should natively support async contexts to comply with Django's async safety guidelines. There should be no warning messages about the asynchronous handler being adapted for the middleware, ensuring optimal performance and compatibility in async environments.

**Environment**

-   Python version: 3.11.0
-   Django version: 4.2.4
-   Django-axes version: 6.1.1
-   Operating system: Mac OS X 13.5.1

**Additional context** 

The lack of native async support could potentially lead to performance issues and unexpected behaviors when the application is operating in async mode. Adhering to Django's recommendations for async safety is crucial to ensure that the middleware performs optimally and is fully compatible with Django's async features.

aleksihakli:
Hey! We're open to pull requests so if someone has the opportunity to look into & implement the async support in the middleware and possible other necessary layers it'd be great.

aleksihakli:
This has been fixed with help from @Taikono-Himazin in version 6.3.0, thank you both for the bug report and the fix!

PR #1147, opened 2023-12-13
Add async support to middleware

Taikono-Himazin:
Added middleware Async support.
I tried using it at hand, and it seems to have no problems. The existing pytest is also passing.

However, I wasn't sure how to add documentation and tests.
I would like to know if you know how to switch between Django's Async mode and Sync mode, and how to run pytest while switching between Async mode and Sync mode.

Related issue
- #1123

### DIFF
--- a/axes/middleware.py
+++ b/axes/middleware.py
@@ -1,5 +1,6 @@
 from typing import Callable
 
+from asgiref.sync import iscoroutinefunction, markcoroutinefunction, sync_to_async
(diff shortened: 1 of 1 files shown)


krystofbe:
I've tested this in my async views and it works like a charm!

aleksihakli:
Thank you for the PR @Taikono-Himazin 

This has been released as part of version 6.3.0

PR #1169, opened 2024-02-15
Remove compatibility with `pkg_resources`

Viicos:
3.8 is the minimal Python version


aleksihakli:
Released as version 6.3.1.

PR #1198, opened 2024-04-30
Add session hash to access log

### LINKED ISSUE #1190: BUG: Concurrent session logout are miss-tracked
**Describe the bug**
When there are concurrent sessions held by the same user on differente devices the first which logs-out also marks the time on the latter.

**To Reproduce**
1. Configure AXES using `AxesDatabaseHandler` (the default one)
1. Create a user
2. Log in with that user using a browser
3. Log in with that user using a different browser (or incognito mode)
4. <repeat step 3 with other browser/client if needed>
5. Log out from one of the active browsers

Now every `AccessLog` for that user has the same `logout_time`, even those for which there is still an active session. It is not possible to update access-logs for those records.
https://github.com/jazzband/django-axes/blob/fd9d185ad32570434d14de79e45976c6e18fcc54/axes/handlers/database.py#L319-L321

**Expected behavior**
Every session for a single user should be related to a single `AccessLog`, to enable a correct tracking of the user.

**Your environment**
python version: 3.10
django version: 4.2
django-axes version: 
Operating system: Linux

**Possible implementation**
It would be advisable to have an other optional field on `AccessLog` which can be a digest of the current session-id.
This could also be used as a method do detect whenever an access "expires" without log-off.
The reason for not using a FK to session is:
- it is not secure since the PK of session usually is the session-id which must be kept secret and hard to find (if possible)
- already `username` is not a real FK to user but just simple column

sevdog:
# What does this PR do?
This PR adds a new field on `AccessLog` to better track sessions.

There is still an issue for sessions in which a user changes its password: because `django.contrib.auth` creates a new session upon password change and no logout/login signal is fired since this is neither a login nor a logout. I suppose it would be fine to keep this case "untracked" since it is not possible to hook into that method and usually changing password is not a frequent task.

When no session engine is configured it falls back to current behaviour: a logout will set the `logout_time` to every `AccessLog` without it. However this could be tolerated since there is no standard way to track such case and usually when there is no session there is not the concept of login/logout.

Fixes #1190 

## Before submitting


axes/helpers.py
```diff
@@ -614,3 +615,24 @@ def inner(*args, **kwargs):  # pylint: disable=inconsistent-return-statements
...
+    """
+    Get client session and returns the SHA256 hash of session key, forcing session creation if required.
+
+    If no session is available on request returns an empty string.
+    """
+    try:
+        session = request.session
+    except AttributeError:
+        # when no session is available just return an empty string
+        return ""
```
aleksihakli, line 630: Pondering on if this should be None or "" but both should work.

axes/migrations/0009_add_session_hash.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="accesslog",
+            name="session_hash",
+            field=models.CharField(
+                blank=True,
+                default="",
+                max_length=64,
```
aleksihakli, line 18: SHA256 should be exactly 64 characters so it might be sensible to force this to either None or 64 characters with e.g. [MinLengthValidator](https://docs.djangoproject.com/en/dev/ref/validators/#minlengthvalidator)

aleksihakli (approved):

aleksihakli:
Released as version 6.5.0.

PR #1209, opened 2024-06-29
fix: make 0007_alter_accessattempt_unique_together.py migration backwards compatible

### LINKED ISSUE #1191: migrations not reversible
During a deployment roll-back, we encountered an irreversible operation exception from django-axes. I had to --fake to rollback 0007

axes/migrations/0007_alter_accessattempt_unique_together.py
line 38 should be
`migrations.RunPython(deduplicate_attempts, reverse_code=migrations.reverse_code=migrations.RunPython.noop),`

The migration removes some duplicates. When rolling back, it's not necessary (or possible) to restore the duplicates, so the backwards migration should just be a noop.

hirotasoshu:
# What does this PR do?

This PR applies fix suggested by @kostrom and fixes #1191

### DIFF
--- a/axes/migrations/0007_alter_accessattempt_unique_together.py
+++ b/axes/migrations/0007_alter_accessattempt_unique_together.py
@@ -35,7 +35,9 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(deduplicate_attempts),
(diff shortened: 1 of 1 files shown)


aleksihakli:
Thank you, looks good!

Issue #667, opened 2020-10-20
Avoid reinitilizing AXES_COOLOFF_TIME for repeat attempts


yahyadev69:
I integrated django-axes to my django application whith the following configs :

15 minutes (as timedelta object for minutes)
AXES_COOLOFF_TIME = timedelta(minutes=15)

Here are the spteps :

1 - I tried 3 time to connect with correct user and wrongs password --> result : the user is locked out (i see the messsage : user locked out during 15 min) 

2 - after 10 minutes, i try again to connect with correct user and password
--> result : the user is locked out again and have to wait another 15 minites ! in total i lost 25 mintes waiting instead to 15 !

Is there a way to avoid this probleme ? I want to use the first duration and not account the second.

Appreciate you help

hramezani:
When the user is locked, every login attempt considered to be a failed login, and every failed login update the `attempt_time` for the locked account. So, `AXES_COOLOFF_TIME` will be calculated after the last attempt time(Whether successful or unsuccessful)

aleksihakli:
Well, it's possible to implement a check for only updating the cool off time if it is not valid anymore, but do you think this should be part of the default behaviour?

jachane99:
On the topics of cooloff time, I'm looking to implement django axes for brute force protection for my application. I want to implement something like this "lockout for 5 minutes after 5 failed attempts, lockout for 10 minutes after another 5 attempts" Seems like the example in the doc is more straightforward with a fixed wait time, so I want to know if it's possible to implement something like this.

aleksihakli:
The incrementing or growing wait time is not supported at the moment, I think, so it would have to be a new feature. The project is open for contributions if you would be interested in making this available for everyone.

jachane99:
Regarding the previous comment, do you have suggestions on how to implement a check for only updating the cool off time? Would it be appropriate to override the `DatabaseDefaultHandler` to update the record while valid?

aleksihakli:
The cool off time is static at the moment, but if you wanted to make it more dynamic, you would have to indeed specialize the database handler for your needs, if you are using the database for tracking failures in your project. At the moment the tracking is done with static attempt time update and a timedelta calculation from there, so you would have to e.g. skip updating the attempt time with certain conditions:

https://github.com/jazzband/django-axes/blob/2d228ea49bea91900b9f94c36d839ed454d17ea5/axes/handlers/database.py#L148

Issue #218, opened 2017-01-17
Implement exponentially growing lockout time


BeOleg:
Example use case:
first lockout in a day: 5 minutes.
2nd: 10 minutes
3rd: 30 minutes
4th: 1 hours
5th: 1 day

camilonova:
PRs are welcome.

PR #1222, opened 2024-08-14
feat: pass request to `AXES_COOLOFF_TIME` callback

browniebroke:
# What does this PR do?

If the `AXES_COOLOFF_TIME` is a callable or path to a callable taking an argument, pass the username to it.

This should enable users to customize the cool off to be user dependant, and possibly implement a growing cool-off time:

- First lockout cools off after 5 mins
- Second one after 10 mins
- etc...

Past references of this:
- https://github.com/jazzband/django-axes/issues/667#issuecomment-737751694
- #218

This PR should be backwards compatible with previous behaviour: if the `AXES_COOLOFF_TIME` is a callable that takes no argument, it will be called without parameters.

## Before submitting


browniebroke:
Looks like the test failures are the same as another PR (https://github.com/jazzband/django-axes/pull/1221) and they don't look related to my change so I'm going to ignore them for this. Any idea how to fix them? I'll try to take a look if I find time

axes/helpers.py
```diff
@@ -43,21 +45,22 @@ def get_cache_timeout() -> Optional[int]:
...
 
-    cool_off = get_cool_off()
+    cool_off = get_cool_off(username)
     if cool_off is None:
         return None
     return int(cool_off.total_seconds())
 
 
-def get_cool_off() -> Optional[timedelta]:
+def get_cool_off(username: Optional[str] = None) -> Optional[timedelta]:
```
aleksihakli, line 54: Would it be better to just supply the `request` object here so the user can use any field present in there? IP address or other metadata could be interesting as well. The request can also be used to carry attached attributes used in upstream / downstream processing in the middleware stack, if needed. `username` seems a bit limiting if we're implementing features here.

Adding the use of `request` object requires a few modifications in other places as well, but should be easy enough to test.

axes/models.py
```diff
@@ -53,7 +53,9 @@ class Meta:
 
 class AccessLog(AccessBase):
     logout_time = models.DateTimeField(_("Logout Time"), null=True, blank=True)
-    session_hash = models.CharField(_("Session key hash (sha256)"), default="", blank=True, max_length=64)
+    session_hash = models.CharField(
```
aleksihakli, line 56: This is an unrelated change, please separate reformatting and stylistic things from feature changes into separate commits at the very least.

aleksihakli (changes requested):
Looks good but I think the implementation should just take the request if a parametrizable callback is implemented as that contains the username and other metadata that is useful for processing as well, offering users wanting to customize more flexibility for the same price.

aleksihakli:
> Looks like the test failures are the same as another PR (#1221) and they don't look related to my change so I'm going to ignore them for this. Any idea how to fix them? I'll try to take a look if I find time

There was a failing test that was produced by a change in the Django upstream implementation that limits header sizes. I updated the test, if you rebase this PR changeset on top of jazzband/django-axes repository master branch the test should be fixed in this branch as well.

axes/helpers.py
browniebroke: That suggestion made a lot of sense when I read it, but now that I'm trying to make the changes, I remembered why I went for this initial implementation... 

The place where it's called, `user_login_failed` calls a function to resolve the username `get_client_username`, passing down some `credentials`. I can see that this function defaults to using `request.data` or `request.POST`.

I think I'm lacking a bit of context on the lifecycle where this is used, is that fair to assume that this will be called in a POST request where the username will be included in the body? It feels we might not be there in some cases like SSO/OAuth flow. Are these kind of flows out of the scope of django-axes?

axes/models.py
browniebroke: Apologies, I've reverted that now.

axes/attempts.py
```diff
@@ -12,17 +12,18 @@
 log = getLogger(__name__)
 
 
-def get_cool_off_threshold(attempt_time: Optional[datetime] = None) -> datetime:
+def get_cool_off_threshold(request: Optional[HttpRequest] = None) -> datetime:
```
browniebroke, line 15: As the `attempt_time` is extracted from the request, I've dropped the first argument and extract it internally. 

While it's not part of the documented API, it would be a potential breaking change, if for some reasons someone uses this function directly.

Should I leave the parameter for backwards compatibility with a warning if a value is passed?

browniebroke:
I've pushed more updates, trying to me more granular with my commits. I didn't rebase anything yet to help you keep track of the changes you already reviewed, and the new ones I've added. I can rebase them when/if you want me to.

aleksihakli (changes requested):
Thanks @browniebroke the changeset looks good 👍 Do you think it would be feasible to drop the `_maybe_partial` implementation and just directly invoke the `settings.AXES_COOLOFF_TIME` callable without the `functools.partial` and `inspect` additions? I think that'd make for an easier-to-understand API in the long run and we can just bake a breaking release for this changeset with e.g. major release version 7 if needed, it's not a big effort. I can write up the changelog and bake a release after this has been merged.

browniebroke:
I think it's all done now. Let me know if I should update the changelog with some "unreleased changes".

axes/helpers.py
aleksihakli: Hey, a good question on the authentication methods and flows, I initially missed this one. Sorry about that.

SSO flows such as SAML and OAuth are absolutely in scope of this plugin. Ideally django-axes should be login method agnostic, if possible.

If anyone else is checking this discussion it might be useful to see the Django authentication signals documentation at https://docs.djangoproject.com/en/dev/ref/contrib/auth/#module-django.contrib.auth.signals for reference.

You are right in that the `request` object might not contain all the relevant information needed to dig up e.g. username or other identifying login data if the login is coming in from e.g. a SAML request, but `request` is the only persistent information, really, that we get from all the login flows i.e. going up and down the middleware stack and in all the login signals as described at e.g. https://django-axes.readthedocs.io/en/latest/7_architecture.html

At the current time we always expect a `request` for the [`get_lockout_response`](https://github.com/jazzband/django-axes/pull/1222/files#diff-87f5ab498b667653f3c1f46b6a38cbe11ec0ffbad6d423a5a91c4055123ba218L445) callable and `credentials` as an optional. 

It actually might be nice to pass the `credentials` argument as well, as you pointed out.

If you're open to it I think it'd be OK to add that argument as well and bake a version 7.x from that. If you want to make a PR I'll be happy to make a new release for that.

aleksihakli:
Hey @browniebroke I made a release version 7.0.0 with associated changelog and notes for upgrading, do let me know if there's anything missing.

I also missed your earlier comment regarding e.g. SSO logins, please check it out and let me know if you have any thoughts on that.

Cheers for the good PR 👍

browniebroke:
Thanks a lot for the quick feedback and very prompt release!

herg:
Love this feature, but if I understand the library correctly it seems like `clean_expired_user_attempts` needs to be refactored for this to work.

The way I read it, `clean_expired_user_attempts` [deletes](https://github.com/jazzband/django-axes/blob/ff9c3296ef84ad9b75309ba4a659320419031d76/axes/attempts.py#L83) all `AccessAttempt`s that are expired based on the *current* request's cooloff threshold, regardless of whether those attempts had different cooloffs.

This means if UserA has an attempt recorded with a 60 minute cooloff, and then UserB comes in 10 minutes later with a cooloff of 5 minutes, we'll delete "all `AccessAttempt`s older than 5 minutes". UserA's `AccessAttempt` will be deleted even though it doesn't expire for another 50 minutes.

Am I understanding this limitation correctly? Thanks for the help!

aleksihakli:
Hey @herg and thanks for chipping in. 

Yeah, it's possible to reset other users' cool-offs, if the developer has specified a user-dependant cool-off to be resolved by a callable in https://github.com/jazzband/django-axes/blob/479a355d2221869a313f1e029ef60a5f3976f7b9/axes/helpers.py#L52.

I've created https://github.com/jazzband/django-axes/issues/1302 for discussing this further.

Would you like to triage and possibly propose a solution for this?

PR #1231, opened 2024-10-02
docs: fix broken settings table

browniebroke:
# What does this PR do?

The settings table is not rendering on this page: https://django-axes.readthedocs.io/en/latest/4_configuration.html

Looks like I broke the rendering of the settings table in my previous PR #1222 🤦🏻  
It's because the closing `|` of the `AXES_COOLOFF_TIME` row is not aligned with the other rows in the table

## Before submitting


aleksihakli:
Thank you, looks good 👍

PR #1252, opened 2024-10-28
fix(docs): AXES_CLIENT_IP_CALLABLE documentation

OscarVanL:
# What does this PR do?

Previously, the readme suggested the AXES_CLIENT_IP_CALLABLE function should have two arguments, but actually it only takes one argument.

## Before submitting


aleksihakli:
Thank you @OscarVanL!

PR #1270, opened 2025-01-17
Use AUTH_USER_MODEL's USERNAME_FIELD as default for setting AXES_USERNAME_FORM_FIELD instead of hardcoded "username"

amneher:
# in /axes/conf.py:47, instead of defaulting to "username" for the value of AXES_USERNAME_FORM_FIELD, get the current user model's USERNAME_FIELD.

Fixes # 1133

## Before submitting
- [X] Did you make sure to update the documentation with your changes?


aleksihakli:
Thank you, looks good!

aleksihakli:
This has been released in version 7.0.2.

ippeiukai:
This is an improvement, but please note this was actually a breaking change.
We use "username" field in the form while AUTH_USER_MODEL's USERNAME_FIELD is customized to something else.

(When upgrading to 7.0.2 or above, if AUTH_USER_MODEL's USERNAME_FIELD is not "username" but AXES_USERNAME_FORM_FIELD was not set, one must set AXES_USERNAME_FORM_FIELD = "username".)

PR #1300, opened 2025-04-11
chore: add Django 5.2 support

mkniewallner:
# What does this PR do?

Add explicit support for Django 5.2, which was [released on April 2, 2025](https://docs.djangoproject.com/en/5.2/releases/5.2/) and test against it on the CI.

Also fixing 2 issues spotted with the current test suite:
- tox `dj42` sets incorrect bounds, so we actually test against Django 4.1, as can be seen [here](https://github.com/jazzband/django-axes/actions/runs/14309463862/job/40100723125#step:7:94):
```console
  py312-dj42: commands[0]> pytest
  ============================= test session starts ==============================
  platform linux -- Python 3.12.9, pytest-8.3.4, pluggy-1.5.0
  cachedir: .tox/py312-dj42/.pytest_cache
  django: version: 4.1.13, settings: tests.settings (from ini)
```
- Python 3.13 + Django 4.2 on the CI actually tests against Django 5.2 (as this is not a possible combination in tox), as can be seen [here](https://github.com/jazzband/django-axes/actions/runs/14309463862/job/40100723171#step:7:92):
```console
  py: commands[0]> pytest
  ============================= test session starts ==============================
  platform linux -- Python 3.13.2, pytest-8.3.4, pluggy-1.5.0
  cachedir: .tox/py/.pytest_cache
  django: version: 5.2, settings: tests.settings (from ini)
```
Alternatively, a fix could be to allow Python 3.13 for Django 4.2 on tox, but since 3.13 is not officially supported for 4.2, not sure if it's worth

Not really related to this PR, but also spotted that support for Python 3.7 + 3.8 and Django 3.2 have been dropped, and mentioned in the changelog, but the library still allows installation on those versions. Made 2 separate commits to handle that as well, but maybe this was done on purpose, in which case tell me, I can remove the last 2 commits.

## Before submitting

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
-        django-version: ['4.2', '5.0', '5.1']
+        django-version: ['4.2', '5.0', '5.1', '5.2']
         include:
           # Tox configuration for QA environment
           - python-version: '3.13'
@@ -29,6 +29,8 @@ jobs:
             django-version: '5.0'
           - python-version: '3.9'
             django-version: '5.1'
+          - python-version: '3.9'
+            django-version: '5.2'
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


aleksihakli:
LGTM, thank you for the PR!

PR #1306, opened 2025-05-03
refactor: move db accessing attempt functions to db handler methods

nefrob:
### What

Follow up from #1305.

- Moves all database accessing functions from `axes.attempts` to methods in `axes.handlers.database` to allow for customization via subclassing

This will be a breaking change for anyone importing these functions from `axes.attempts`.

### DIFF
--- a/axes/attempts.py
+++ b/axes/attempts.py
@@ -1,13 +1,10 @@
 from logging import getLogger
-from typing import List, Optional
+from typing import Optional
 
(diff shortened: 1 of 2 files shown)


aleksihakli (approved):
Thank you! Looks good, all tests passing, I'll go ahead and merge this 👍

aleksihakli:
This has been released as part of django-axes version 8.0.0 along with #1305

PR #1313, opened 2025-05-27
Added individual attempt expiry feature for #1302

### LINKED ISSUE #1302: User-dependant cool-off can reset attempts related to other users
> Love this feature, but if I understand the library correctly it seems like `clean_expired_user_attempts` needs to be refactored for this to work.
> 
> The way I read it, `clean_expired_user_attempts` [deletes](https://github.com/jazzband/django-axes/blob/ff9c3296ef84ad9b75309ba4a659320419031d76/axes/attempts.py#L83) all `AccessAttempt`s that are expired based on the *current* request's cooloff threshold, regardless of whether those attempts had different cooloffs.
> 
> This means if UserA has an attempt recorded with a 60 minute cooloff, and then UserB comes in 10 minutes later with a cooloff of 5 minutes, we'll delete "all `AccessAttempt`s older than 5 minutes". UserA's `AccessAttempt` will be deleted even though it doesn't expire for another 50 minutes.
> 
> Am I understanding this limitation correctly? Thanks for the help!

_Originally posted by @herg in https://github.com/jazzband/django-axes/issues/1222#issuecomment-2802792662_

kuldeepkhatke:
# Individual attempt based longest expiry attempt restriction feature

Implemented a `AXES_INDIVIDUAL_ATTEMPT_EXPIRY` flag based feature, 
such that if we enable it then it will handle attempts cleanup based on the longest expiry of the attempt. 
Hence added expire_at field to implement this feature.

All testcases passed after implementation.

Fixes #1302


aleksihakli:
Comparing to #1310 this adds state to the database which is only used if the flag is switched on, otherwise growing the database table size and seldom using the expiration times. I would separate the new expiration time field into it's own table if it's not used often, and only write it to the database when necessary.

> # Individual attempt based longest expiry attempt restriction feature
> 
> Implemented a `AXES_INDIVIDUAL_ATTEMPT_EXPIRY` flag based feature, such that if we enable it then it will handle attempts cleanup based on the longest expiry of the attempt. Hence added expire_at field to implement this feature.
> 
> All testcases passed after implementation.
> 
> Fixes #1302

Existing test cases pass, but there are no test cases for the new flag / feature, which need to be added if this is to be merged upstream.

axes/conf.py
```diff
@@ -87,6 +87,8 @@
 
 settings.AXES_COOLOFF_TIME = getattr(settings, "AXES_COOLOFF_TIME", None)
 
+settings.AXES_INDIVIDUAL_ATTEMPT_EXPIRY = getattr(settings, "AXES_INDIVIDUAL_ATTEMPT_EXPIRY", False)
```
aleksihakli, line 90: Maybe `AXES_USE_ATTEMPT_EXPIRATION`?

The cache backend also has a TTL for the entries: https://github.com/jazzband/django-axes/blob/master/axes/handlers/cache.py#L116

I'm wondering if this should reflect the database handlers' default implementation?

kuldeepkhatke: We can modify the name to `AXES_USE_ATTEMPT_EXPIRATION` . Its LGTM. 
Seems more explanatory. 

Instead of cache TTL, i'll better prefer timestamp data. To avoid calculations at every clean_expired_user_attempts call.

axes/migrations/0010_accessattempt_expires_at.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('axes', '0009_add_session_hash'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='accessattempt',
+            name='expires_at',
+            field=models.DateTimeField(blank=True, help_text='The time when this access attempt expires and is no longer valid.', null=True, verbose_name='Expires At'),
```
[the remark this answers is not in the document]
kuldeepkhatke: Sure I'll create a separate `AccessAttemptExpiration` table in a OneToOne connection.

Such that a record will create only when the flag is enabled.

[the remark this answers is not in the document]
kuldeepkhatke: If we talk about timestamp vs timedelta thing. then my answer is same. 

Instead of cache TTL, i'll better prefer timestamp data. To avoid calculations at every `clean_expired_user_attempts()` call.

kuldeepkhatke (commented):
I'm modifying the changes as per suggestions, & will add test cases accordingly.

kuldeepkhatke:
Hey @aleksihakli , 
I've made the required changes as per suggestion. 

Added testcases for the feature. 

Please review the changes.

kuldeepkhatke:
Also, i'm curious if we can consider adding a AUTHORS file to add the name of list of contributors

aleksihakli:
> Hey @aleksihakli , I've made the required changes as per suggestion.
> 
> Added testcases for the feature.
> 
> Please review the changes.

Looks good, nice job 👍  I made a few suggestions for further improvements but I think this is close to merge ready.

axes/attempts.py
```diff
@@ -24,3 +24,19 @@ def get_cool_off_threshold(request: Optional[HttpRequest] = None) -> datetime:
     if attempt_time is None:
         return now() - cool_off
     return attempt_time - cool_off
+
+def get_individual_attempt_expiry(request: Optional[HttpRequest] = None) -> datetime:
```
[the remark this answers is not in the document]
kuldeepkhatke: I think its better to keep in axes.helpers.py as this can used at other places also. 

But yes, as per current scope on database handler is the consumer. 
So, you take the call, i'll modify accordingly.

axes/handlers/database.py
```diff
@@ -219,6 +219,23 @@ def user_login_failed(self, sender, credentials: dict, request=None, **kwargs):
...
                     )
 
+                # Set the expiration time for the attempt based on the cool off period.
+                if settings.AXES_USE_ATTEMPT_EXPIRATION:
+                    if not hasattr(attempt, "expiration") or attempt.expiration is None:
+                        log.debug(
+                            "AXES: Creating new AccessAttemptExpiration for %s", client_str
+                        )
+                        # Create a new AccessAttemptExpiration if it does not exist
+                        attempt.expiration = AccessAttemptExpiration(
```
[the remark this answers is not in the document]
kuldeepkhatke: Actually i avoided create function here, as calling `attempt.expiration.save()` in common outside of if else condition.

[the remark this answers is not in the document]
kuldeepkhatke: Keeping option B for now

axes/attempts.py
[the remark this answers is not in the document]
kuldeepkhatke: shifted the function to helpers.py and renamed it

tests/test_handlers.py
```diff
@@ -567,3 +569,61 @@ def test_handler_is_allowed(self):
...
+    @patch("axes.handlers.database.get_cool_off_threshold")
+    @patch("axes.models.AccessAttempt.objects.filter")
+    def test_clean_expired_user_attempts_expiration_false(self, mock_filter, mock_get_threshold, mock_log):
+        mock_get_threshold.return_value = "fake-threshold"
+        mock_qs = MagicMock()
+        mock_filter.return_value = mock_qs
+        mock_qs.delete.return_value = (2, None)
+
+        count = self.handler.clean_expired_user_attempts(request=self.mock_request, credentials=None)
+        mock_filter.assert_called_once_with(attempt_time__lt="fake-threshold")
```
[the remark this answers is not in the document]
kuldeepkhatke: added a new testcase with actual queryset calls

kuldeepkhatke:
Hi @aleksihakli , 
Updated the code as per suggestions, rebased with latest code. 

Thanks

tests/test_handlers.py
```diff
@@ -567,3 +569,82 @@ def test_handler_is_allowed(self):
 
     def test_handler_get_failures(self):
         self.assertEqual(0, AxesProxyHandler.get_failures(self.request, {}))
+
+
+@override_settings(AXES_HANDLER="axes.handlers.database.AxesDatabaseHandler", AXES_COOLOFF_TIME=timezone.timedelta(seconds=10))
+class AxesDatabaseHandlerExpirationFlagTestCase(AxesTestCase):
```
aleksihakli, line 575: Please add test cases for:

1. deleting attempts where no attempts are actually deleted because they are not yet expired; check that the database contains the same attempts before and after deletion attempt
2. some expired attempts are deleted but otherst that have not yet expired are not; check attempts before and after the deletion
3. all attempts are deleted because they are expired

This covers no deletions, partial deletions and complete deletions cases in a way that should catch errors when someone inevitably changes the cleanup logic for expired attempts in the future.

At the moment the tests are checking that *internal* methods are being called for e.g. `log` and the `count` returned by `clean_expired_user_attempts` *looks* good there are not tests for the actual contents of the database before and after running the cleanup methods.

axes/models.py
```diff
@@ -51,6 +51,25 @@ class Meta:
...
+    access_attempt = models.OneToOneField(
+        AccessAttempt,
+        primary_key=True,
+        on_delete=models.CASCADE,
+        related_name="expiration",
+        verbose_name=_("Access Attempt"),
+    )
+    expires_at = models.DateTimeField(
+        _("Expires At"),
+        null=True,
```
aleksihakli, line 64: These do not need to be null or blank as these rows should only ever be created in a case where the expirations are on, right?

axes/handlers/database.py
```diff
@@ -382,13 +398,22 @@ def clean_expired_user_attempts(
...
-        threshold = get_cool_off_threshold(request)
-        count, _ = AccessAttempt.objects.filter(attempt_time__lt=threshold).delete()
-        log.info(
-            "AXES: Cleaned up %s expired access attempts from database that were older than %s",
-            count,
-            threshold,
-        )
+        if settings.AXES_USE_ATTEMPT_EXPIRATION:
+            threshold = timezone.now()
+            count, _ = AccessAttempt.objects.filter(expiration__expires_at__lt=threshold).delete()
```
aleksihakli, line 403: Nitpick: if this is named as `expires_at` and not `expires_after` then the comparison should probably be `expiration__expires_at__lte` instead of `expiration__expires_at__lt` since "at" is exact and includes the moment in time itself; think of "expires at *exactly* 12:00" or ">= 12:00" vs. "expires *exactly* after 12:00" or "> 12:00".

axes/handlers/database.py
```diff
@@ -382,13 +398,22 @@ def clean_expired_user_attempts(
...
+            log.info(
+                "AXES: Cleaned up %s expired access attempts from database that expiry were older than %s",
+                count,
+                threshold,
+            )
+        else:
+            threshold = get_cool_off_threshold(request)
+            count, _ = AccessAttempt.objects.filter(attempt_time__lt=threshold).delete()
```
aleksihakli, line 411: Nitpick: If we use `lte` above then this should also be later-than-or-equal.

axes/admin.py
```diff
@@ -38,11 +49,16 @@ class AccessAttemptAdmin(admin.ModelAdmin):
...
         "post_data",
         "failures_since_start",
+        "expiration",
     ]
 
     def has_add_permission(self, request: HttpRequest) -> bool:
         return False
 
+    def expiration(self, obj: AccessAttempt):
+        if hasattr(obj, "expiration") and obj.expiration.expires_at:
```
aleksihakli, line 59: Maybe something like:

```suggestion
        return obj.expiration.expires_at if hasattr(obj, "expiration") else _("Not set")
```

axes/handlers/database.py
```diff
@@ -219,6 +220,21 @@ def user_login_failed(self, sender, credentials: dict, request=None, **kwargs):
                         client_str,
                     )
 
+                if settings.AXES_USE_ATTEMPT_EXPIRATION:
+                    if not hasattr(attempt, "expiration") or attempt.expiration is None:
```
aleksihakli, line 224: Maybe just create the object if it doesn't exist? 

Is the expiration timestamp ever *earlier* than it previously was? Why should it only be updated to a later value if the `get_attempt_expiration` returns an earlier value?

```suggestion
                if settings.AXES_USE_ATTEMPT_EXPIRATION:
                    expiration, created = AccessAttemptExpiration.objects.update_or_create(
                            access_attempt=attempt,
                            defaults={
                                "expires_at": get_attempt_expiration(request),
                            }
                        )
                    if created:
                        log.debug("AXES: Created new AccessAttemptExpiration for %s", client_str)
                    else:
                        log.debug("AXES: Updated existing AccessAttemptExpiration for %s", client_str)           
```

aleksihakli (changes requested):

axes/handlers/database.py
kuldeepkhatke: ya, make sense

axes/handlers/database.py
kuldeepkhatke: sure, will update

axes/admin.py
kuldeepkhatke: ya, in case of null=False for expires_at field, no more need to have a null check condition

blockchainGuru1018:
approved

axes/helpers.py
```diff
@@ -100,6 +100,21 @@ def get_cool_off_iso8601(delta: timedelta) -> str:
         return f"P{days_str}T{time_str}"
     return f"P{days_str}"
 
+def get_attempt_expiration(request: Optional[HttpRequest] = None) -> datetime:
```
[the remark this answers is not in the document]
kuldeepkhatke: This function have separate usecase than get_cool_off_threshold() function. 

get_attempt_expiration() get call at the time of attempt executed. 

while get_cool_off_threshold()  get called inside the clean_expired_user_attempts() function. 

Also get_attempt_expiration() works on adding the cool_off time 

while  get_cool_off_threshold() works on subtracting the cool_off time.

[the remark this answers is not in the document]
aleksihakli: That's true, I was wondering if they should be co-located in the same file since they have very similar logic.

kuldeepkhatke:
Absolutely, No Problem @aleksihakli , 

Thanks for appreciation & merging my PR.

PR #1322, opened 2025-06-16
Locked out status feature

### LINKED ISSUE #1320: FEATURE REQUEST: Attempt Status feature with filter feature
**Is your feature request related to a problem? Please describe.**
It is not easy to get which records are locked out and which still have some attempts remaining.

**Describe the solution you'd like**
It would be better to have a status column for the **AccessAttemptAdmin** list view, such that we can quickly identify the locked out records & also find out how many attempts remaining for the non locked out records. 

**Additional context**
We also have a is_locked_out filter based on this status, such that a specific set of records can be filtered.

kuldeepkhatke:
# What does this PR do?
Closes #1320

Added a status feature to the **AccessAttemptAdmin** list view
which is not a field but just a calulation based on settings.AXES_FAILURE_LIMIT and attempt.failures_since_start

image

Also added IsLockedOutFilter option

image

### DIFF
--- a/axes/admin.py
+++ b/axes/admin.py
@@ -21,6 +21,11 @@ class AccessAttemptAdmin(admin.ModelAdmin):
 
     list_filter = ["attempt_time", "path_info"]
 
+    if isinstance(settings.AXES_FAILURE_LIMIT, int) and settings.AXES_FAILURE_LIMIT > 0:
+        # This will only add the status field if AXES_FAILURE_LIMIT is set to a positive integer
+        # Because callable failure limit requires scope of request object
+        list_display.append("status")
+
     search_fields = ["ip_address", "username", "user_agent", "path_info"]
 
     date_hierarchy = "attempt_time"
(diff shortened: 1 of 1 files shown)


kuldeepkhatke:
Hi @aleksihakli , please review this mr

axes/admin.py
```diff
@@ -6,18 +6,42 @@
...
+        if self.value() == "yes":
+            return queryset.filter(failures_since_start__gte=settings.AXES_FAILURE_LIMIT)
+        elif self.value() == "no":
+            return queryset.filter(failures_since_start__lt=settings.AXES_FAILURE_LIMIT)
+        return queryset
+
+
 class AccessAttemptAdmin(admin.ModelAdmin):
-    list_display = (
+    list_display = [
```
aleksihakli, line 28: Nit: style changes should be split into a separate commits for clarity.

axes/admin.py
```diff
@@ -43,6 +67,14 @@ class AccessAttemptAdmin(admin.ModelAdmin):
     def has_add_permission(self, request: HttpRequest) -> bool:
         return False
 
+    def status(self, obj: AccessAttempt):
```
aleksihakli, line 70: This method first checks for locked out status and then checks for it again inside a reduntant inlined if statement, needs a cleanup.

aleksihakli (changes requested):

axes/admin.py
kuldeepkhatke: okay, will split the commits

axes/admin.py
kuldeepkhatke: but its too much to be handled in inlined if, i'll better break it to standard if else statement

kuldeepkhatke:
Updated changes, as requested @aleksihakli

aleksihakli:
@kuldeepkhatke would you have the opportunity of rebasing this on top of the current master branch and #1313?

kuldeepkhatke:
Rebased the mr, please review once @aleksihakli

aleksihakli:
Thanks, I’ll go ahead and merge this 👍

PR #1401, opened 2026-02-21
feat: Add `Retry-After` HTTP header to lockout responses

rodrigobnogueira:
# What does this PR do?

When `AXES_COOLOFF_TIME` is configured, lockout responses now automatically include the `Retry-After` HTTP header (RFC 7231 §7.1.3) with the cool-off duration in seconds. This helps clients (browsers, API consumers, bots)
understand how long to wait before retrying.

### `axes/helpers.py`
- Added `_set_retry_after_header()` helper that converts `AXES_COOLOFF_TIME`
  to total seconds and sets the `Retry-After` header on the response.
- Called from the three Axes-controlled branches of `get_lockout_response()`:
  JSON (XHR), template-rendered, and plain `HttpResponse`.

### `tests/test_helpers.py`
- Added 5 tests covering:
  - `Retry-After` present with a valid cool-off time (plain response)
  - `Retry-After` absent when `AXES_COOLOFF_TIME` is `None`
  - `Retry-After` present on JSON (XHR) responses
  - `Retry-After` present on template-rendered responses
  - `Retry-After` absent on redirect responses (`AXES_LOCKOUT_URL`)

### `docs/4_configuration.rst`
- Added a `.. note::` block after the settings table documenting the
  `Retry-After` header behavior and which response types include it.

## Design Decisions

| Scenario                  | Header set? | Rationale                                  |
|---------------------------|-------------|--------------------------------------------|
| Plain `HttpResponse`      | ✅          | Standard lockout response                  |
| `JsonResponse` (XHR)      | ✅          | API consumers benefit most                 |
| Template render            | ✅          | Browser can still use it                   |
| `AXES_LOCKOUT_URL`        | ❌          | Redirect — destination controls headers    |
| `AXES_LOCKOUT_CALLABLE`   | ❌          | User owns the response entirely            |
| `AXES_COOLOFF_TIME=None`  | ❌          | Permanent ban — no retry window to advertise |

## Testing

All 351 existing + new tests pass with zero failures.

## Before submitting


aleksihakli (changes requested):

axes/middleware.py
```diff
@@ -39,6 +39,22 @@ def __init__(self, get_response: Callable) -> None:
...
+
+        if settings.AXES_LOCKOUT_CALLABLE or settings.AXES_LOCKOUT_URL:
+            return response
+
+        cool_off = get_cool_off(request)
+        if cool_off is not None:
+            response["Retry-After"] = str(int(cool_off.total_seconds()))
+
+        return response
+
```
aleksihakli: Wouldl it make sense to just write this as a simpler single if-then method? Why do we need to skip setting the header on callable and lockout URLs?

```suggestion
    @staticmethod
    def set_retry_after_header(
        response: HttpResponse, request: HttpRequest
    ) -> HttpResponse:
        if settings.AXES_ENABLE_RETRY_AFTER_HEADER:
            response["Retry-After"] = str(int(get_cool_off(request).total_seconds()))
        return response
```

aleksihakli (changes requested):

axes/middleware.py
rodrigobnogueira: done. It is simpler now. thanks

rodrigobnogueira: Let me know if the implementation is better now.

rodrigobnogueira:
Thanks @aleksihakli — reworked exactly along these lines:

- **Out of the middleware, into `axes.helpers`.** There's now a small `set_retry_after_header(request, response)` invoked from `get_lockout_response`, so the header is set in one place for every built-in lockout response (JSON, template, redirect, default). The middleware is back to untouched upstream, and there's no more callable/lockout-URL special-casing — it just sets the header and mutates the response in place, as you suggested.
- **Toggle flag.** Gated behind `AXES_ENABLE_RETRY_AFTER_HEADER` (default `False`), so current behaviour is unchanged unless opted in.
- A custom `AXES_LOCKOUT_CALLABLE` owns its whole response, so it's left alone; a permanent lockout (no cool-off) sets no header.

Tests now live in `test_helpers.py` against `get_lockout_response` / `set_retry_after_header` directly. Rebased on latest master; `mypy`, `prospector`, and `black` on `axes/` are green. Happy to also tag the callable path if you'd prefer.

aleksihakli:
Hey @rodrigobnogueira, thank you for the PR and sorry for the late reply. I merged a couple of PRs and there was a conflict in this one - could you rebase this on top of the latest master and let's get this merged then?

aleksihakli:
LGTM, thank you!

PR #1440, opened 2026-05-18
docs: fix misaligned column padding in configuration settings table

oniku-mogumogu:
Fixes a rendering failure in the configuration settings table in `docs/4_configuration.rst`.

Due to misaligned column boundaries in the RST grid table, Docutils/Sphinx was failing to parse the table during the HTML build. As a result, the entire table was not rendered at all in the browser, leaving the configuration reference missing from the page.

The root cause was four data rows where the trailing pipe character (`|`) was not positioned at the column boundary defined by the separator rows (`+---+`). Docutils requires exact alignment between `|` in data rows and `+` in separator rows; any deviation causes the entire table block to be rejected by the parser.

Affected rows and adjustments:
- `AXES_FAILURE_LIMIT` (line 22): added 2 spaces before closing `|`
- `AXES_COOLOFF_TIME` (line 26): removed 2 spaces before closing `|`
- `AXES_USE_ATTEMPT_EXPIRATION` (line 28): added 54 spaces before closing `|`
- `AXES_LOCKOUT_CALLABLE` (line 58): removed 1 space before closing `|`

No content changes were made — only whitespace padding was adjusted to restore correct table parsing and proper HTML rendering.

## Before submitting


aleksihakli:
Thank you @oniku-mogumogu ! The updated documentation should be available on RTD momentarily.

PR #1458, opened 2026-07-03
Remove unused imports and variables (F401, F841)

Mubashir78:
Remove unused imports and dead variable assignments flagged by Ruff rules F401 and F841.

### DIFF
--- a/axes/apps.py
+++ b/axes/apps.py
@@ -30,6 +30,11 @@ class AppConfig(apps.AppConfig):
         # Only import settings, checks, and signals one time after Django has been initialized
         from axes.conf import settings
 
+        # Side-effect imports: checks register system checks via @register,
+        # signals register signal handlers at import time.
+        from axes import checks as _checks  # noqa: F401
+        from axes import signals as _signals  # noqa: F401
+
         # Skip startup log messages if Axes is not set to verbose
         if settings.AXES_VERBOSE:
             if callable(settings.AXES_LOCKOUT_PARAMETERS) or isinstance(
--- a/axes/apps.py
+++ b/axes/apps.py
@@ -29,7 +29,6 @@ class AppConfig(apps.AppConfig):
 
         # Only import settings, checks, and signals one time after Django has been initialized
         from axes.conf import settings
-        from axes import checks, signals
 
         # Skip startup log messages if Axes is not set to verbose
         if settings.AXES_VERBOSE:
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -1,4 +1,3 @@
-from platform import python_implementation
 from unittest.mock import MagicMock, patch
 from datetime import datetime, timezone as dt_timezone
 from django.test import override_settings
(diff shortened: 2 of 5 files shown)


axes/apps.py
```diff
@@ -29,7 +29,6 @@ def initialize(cls):
 
         # Only import settings, checks, and signals one time after Django has been initialized
         from axes.conf import settings
-        from axes import checks, signals
```
stereodamage, line 32: I'm not quite sure with this removal - it's a side-effect import. `checks.py` registers 6 system checks via `@register` at import time and nothing else imports that module, so after your change `manage.py check` runs zero axes checks:

```python
>>> [c.__name__ for c in registry.registered_checks if 'axes' in c.__module__]
[]
```

`test_checks.py` still passes because the test file imports `axes.checks` itself, which re-registers them as a side effect. 

The signals half survives mostly by accident: `axes/admin.py` imports the database handler, which imports `axes.signals`, so admin autodiscovery connects the receivers at startup. Even without the admin app, the first `authenticate()` call lazily imports the configured handler (backends.py calls `AxesProxyHandler.is_allowed` before `user_login_failed` fires at the end of the same call), and both built-in handlers import `axes.signals`. But that's the built-in handlers happening to import the right module. A project running a custom `AXES_HANDLER` (subclassing `AxesHandler`, like the docs suggest) without `django.contrib.admin` ends up with zero receivers and its `user_login_failed` is never called. Reproduced on this branch: failed login -> handler never notified; restore the import -> notified.

The file starts with `# pylint: disable=unused-import` for exactly this line, so I'd keep it and silence Ruff with `# noqa: F401` instead.

stereodamage (changes requested):

Mubashir78:
Thanks for the thorough review! I've reverted the removal of `from axes import checks` and `from axes import signals` in `axes/apps.py`, adding `# noqa: F401` markers with `as _checks` / `as _signals` aliases to suppress Ruff while keeping the side-effect imports.

stereodamage:
Thanks! Looks good to me. @aleksihakli ?

Mubashir78:
@stereodamage @aleksihakli Just wanted to gently follow up on this.

axes/apps.py
aleksihakli: Yeah this is a usually less noticed feature of the Python imports system, it can and does run code on import and as such needs to be paid attention to when handling these kinds of dynamic execution and "registration" behaviours. Django has some sugar coating on top and introduces a few extras as well :)

tests/test_handlers.py
```diff
@@ -324,7 +322,7 @@ def test_user_login_failed_only_user_failures_with_none_username(self, log):
         credentials = {"username": None, "password": "test"}
         sender = MagicMock()
         AxesProxyHandler.user_login_failed(sender, credentials, self.request)
-        attempt = AccessAttempt.objects.all()
+        AccessAttempt.objects.all()
```
aleksihakli, line 325: This might be an obsolete line of code altogether.

tests/test_handlers.py
```diff
@@ -334,7 +332,7 @@ def test_user_login_failed_with_none_username(self):
...
```
aleksihakli, line 335: And here as well.

aleksihakli (approved):

PR #1489, opened 2026-08-28
Handle a cache key expiring between add and incr

### LINKED ISSUE #1144: BUG: Race condition using redis with cooloff enabled
**Describe the bug**

After updating axes to v6+, these errors started to pop up every now and then:

```
    AxesProxyHandler.user_login_failed(sender=None, request=request, credentials=credentials)
  File "axes/helpers.py", line 614, in inner
    return func(*args, **kwargs)
  File "axes/handlers/proxy.py", line 119, in user_login_failed
    return cls.get_implementation().user_login_failed(
 File "axes/handlers/cache.py", line 123, in user_login_failed
    failures.append(self.cache.incr(key=cache_key, delta=1))
  File "django_redis/cache.py", line 32, in _decorator
    return method(self, *args, **kwargs)
  File "django_redis/cache.py", line 116, in incr
    return self.client.incr(*args, **kwargs)
  File "django_redis/client/default.py", line 428, in incr
    return self._incr(key=key, delta=delta, version=version, client=client, ignore_key_check=ignore_key_check)
  File "django_redis/client/default.py", line 400, in _incr
    raise ValueError("Key '%s' not found" % key)
```

`AxesProxyHandler.user_login_failed` is called manually in an authentication layer for api endpoints in this case. It's likely this wil increase the chances of this race condition to happen when users are making api calls at a fixed interval.

**To Reproduce**

1. set a AXES_COOLOFF_TIME, eg. 15 minutes
2. on an empty cache, do a first request which you know will be unauthenticated
3. repeat the request exactly AXES_COOLOFF_TIME later
4. observe the ValueError from the stacktrace above.

Since I believe this to be a race condition, this is difficult to reproduce naturally. I think there is always a risk of this happening when axes is configured with a cooloff time. I have only tested with redis cache, I don't know if other backends are affected as well.
In `user_login_failed` the method `self.cache.add` is called first, followed by `self.cache.incr` - if the `added` value is False. Source code snippet:

```
        for cache_key in cache_keys:
            added = self.cache.add(key=cache_key, value=1, timeout=cache_timeout)
            if added:
                failures.append(1)
            else:
                failures.append(self.cache.incr(key=cache_key, delta=1))
                self.cache.touch(key=cache_key, timeout=cache_timeout)
```

Because these two operations are not atomic/in a transaction together, it should be possible for the key to expire _after_ `self.cache.add` and _before_ `self.cache.incr`.

To always get this error on the second request I modified `user_login_failed` by expiring the key before the call to`incr`:

```
        for cache_key in cache_keys:
            added = self.cache.add(key=cache_key, value=1, timeout=cache_timeout)
            if added:
                failures.append(1)
            else:
                self.cache.expire(cache_key, timeout=-1)  # flush key
                failures.append(self.cache.incr(key=cache_key, delta=1))
                self.cache.touch(key=cache_key, timeout=cache_timeout)
```

**Expected behavior**

I expect django-axes to not cause ValueErrors.

**Your environment**
python version: 3.9
django version: 3.2.23
django-axes version: 6.1.1
Operating system: debian bullseye

**Additional context**

This system is running Django 3.x and uses django_redis as the caching backend. That package has an additional flag for incr() that skips the exists-check and avoids raising a ValueError: `ignore_key_check` found here https://github.com/jazzband/django-redis/blob/d7f1292d60ba9d2176149a9ed46b1e1caffb3c0b/django_redis/client/default.py#L557

Even with the call to `self.cache.expire` passing `ignore_key_check=True` no longer raises any ValueErrors:

```
failures.append(self.cache.incr(key=cache_key, delta=1, ignore_key_check=True))
```

Since this is a flag specific to django_redis I did not create a merge request. 

**Possible implementation**

-

aviseth:
Fixes #1144.

`AxesCacheHandler.user_login_failed` adds a failure key, and increments it when the key already exists:

```python
added = self.cache.add(key=cache_key, value=1, timeout=cache_timeout)
if added:
    failures.append(1)
else:
    failures.append(self.cache.incr(key=cache_key, delta=1))
```

Those are two separate cache operations with no atomicity between them, so the key can expire after the `add` and before the `incr` — which is what happens when a login failure lands just as the cool off window closes. `@cpoppema` reported it against django-redis, but it is not backend specific: `BaseCache.incr` is documented to raise `ValueError` for a missing key, so every backend can hit it. The result is a `ValueError` propagating out of the login failure path instead of the failure being recorded.

This takes the approach @aleksihakli suggested on the issue — catch the `ValueError` — and re-seeds the key at 1 rather than swallowing it. If the key expired then the cool off period has genuinely elapsed, so this failure starts a new count, which is the same state the `added` branch produces. The re-add carries `cache_timeout`, so `touch` is not needed on that path.

I preferred this to the `ignore_key_check=True` flag the reporter mentioned, since that is specific to django-redis and would not compile against other backends.

The race remains theoretically possible in the sense that two workers can both re-seed, but the count is then off by one rather than raising, and closing that properly would need a backend-level atomic operation the cache API does not expose.

Added a regression test that patches `incr` to raise. It fails on `master` with the original `ValueError` and passes with this change.

`pytest` is green (377 passed, 19 subtests) and `prospector` reports 0 messages. `mypy .` reports one error in `tests/test_middleware.py:10`, which is present on an unmodified checkout and unrelated to this change.

I have not touched `CHANGES.rst` — every entry there landed in a `Version x.y.z` release commit, so it looked like a maintainer-owned file. Happy to add an entry if you would rather contributors do that.


aleksihakli:
Thank you @aviseth!

aviseth:
Probably was to change the branch name from master to main :)

Regards
Avi Seth
(he/him/his)

